PR #41, opened 2026-04-25
feat(capture-broker): Unix socket server + redactor + banner DOM-removal

### LINKED ISSUE #32: feat(capture-broker): Unix socket server + redactor + banner DOM-removal
## Sprint 2 — Capture #1: Broker server + redactor + banner removal

**Spec:** §5.13 (Capture Broker design, v0.1 Unix-domain socket transport — `/run/willbuy/broker.sock`, mode 0660, schema validation, byte caps as defense-in-depth, single-shot typed request per capture), §5.9 (data classification + redaction — PII / secrets redacted before persistence; labeled-context boundary rule), §2 #2 (banner DOM-removal via curated selector list — DO NOT click; remove from DOM).

**Depends on:** S2-2 (writes to `page_captures`), S2-1 (storage backend — Supabase Storage). Independent of `apps/capture-worker` (#8 already merged).

**Outcome:** The Capture Broker server runs as a systemd service on the VM, accepts ONE typed request per capture from a worker connected via Unix socket, validates shape + size, redacts before writing to object storage, persists the `page_captures` row, and returns an ack. Worker has no other channel to reach the host.

## In scope

- `apps/capture-broker/` — Node TS project, listens on `/run/willbuy/broker.sock`.
- Schema (zod) per spec §5.13:
  ```ts
  {
    status: 'ok' | 'blocked' | 'error',
    a11y_tree_b64: string,           // ≤ 10 MB after decode
    screenshot_b64?: string,         // only when study.screenshots_enabled AND after OCR scrub
    banner_selectors_matched: string[],
    overlays_unknown_present: boolean,
    blocked_reason?: string,
    host_count: number,
    breach_reason?: string
  }
  ```
- Connection lifecycle: accept connection → read ONE message (length-prefixed framing) → schema-parse → byte-cap-enforce (a11y_tree decoded ≤ 10 MB; screenshot decoded ≤ 5 MB) → reject with structured error if oversized → on accept: redactor pass → upload artifacts to object storage → INSERT `page_captures` row → write ack → close.
- Redactor (`apps/capture-broker/src/redactor.ts`) per spec §5.9:
  - Detect: `sk-`, `xoxb-`, `AKIA`, `ghp_`, `glpat-`, JWTs, emails, labeled-context hex (32+ chars).
  - DO NOT match: 40-char asset-URL hash, 64-char CSP nonce, 64-char SKU, commit SHA, cache-busted image URL — implement labeled-context boundary rule per spec.
  - Run BEFORE persistence.
- Banner DOM-removal selector list (`configs/banner-selectors.yaml`) — curated list of common cookie banner / GDPR overlay selectors; loaded at broker startup; broker forwards the list to capture-worker (or capture-worker reads same file). Per spec §2 #2: REMOVE from DOM, never click.
- Systemd unit + bind-mount config: socket at `/run/willbuy/broker.sock`, owner root, group `willbuy`, mode 0660. Worker container mounts read-only via Docker volume.

## TDD acceptance

`apps/capture-broker/test/`:

1. Round-trip: send a valid request → broker stores artifacts + writes row → ack returned. Verify storage object exists; verify row matches.
2. Oversized a11y_tree (e.g. 15 MB decoded) → broker rejects with structured error, no storage write, no row.
3. Malformed JSON → broker rejects.
4. Multiple messages on one connection → only the first is accepted; subsequent get rejected (single-shot per spec §5.13).
5. Redactor positive fixtures (`sk-abcd...` etc.) → redacted in stored artifact.
6. Redactor false-positive fixtures (40-char asset hash, 64-char CSP nonce, etc.) → NOT redacted.
7. Banner selectors: a fixture HTML with a known banner selector → curated list matches; spec §2 #2 says capture-worker DOM-removes it (the broker just forwards the selector list).

## Out of scope

- Network namespace + iptables egress firewall around the worker container (S2-9).
- Screenshot OCR scrub (Sprint 3 — opt-in feature per spec §5.17).
- Capture-worker integration code (touches `apps/capture-worker/` — open a thin follow-up PR after this lands; the broker contract is stable so worker integration is mechanical).

## Definition of done

Standard PR contract.

NikolayS:
Closes #32.

## Summary

Spec §5.13 (Capture Broker design — v0.1 Unix-domain-socket transport at \`/run/willbuy/broker.sock\`, mode 0660, schema validation via zod, byte caps as defense-in-depth, single-shot typed request per capture), §5.9 (data classification + redaction with the labeled-context boundary rule that distinguishes labeled secrets from bare 32+ char hex blobs in asset URLs / commit SHAs / SKUs / CSP nonces), §2 #2 (banner DOM-removal — broker loads the curated selector list from \`configs/banner-selectors.yaml\`; capture-worker removes from DOM, never clicks).

This PR is intentionally scoped to the broker server, redactor, and selector list. Capture-worker integration is a thin follow-up (the wire schema is stable). Storage + DB are injected via in-memory test doubles — CI never hits real Supabase or Postgres, per the issue coordination note.

## What's in scope

- \`apps/capture-broker/\` — new TS package. Listens on a Unix domain socket, accepts ONE typed message per connection (length-prefixed framing), validates shape via zod, enforces decoded byte caps (a11y_tree ≤ 10 MiB, screenshot ≤ 5 MiB), redacts artifact body, persists to object storage + \`page_captures\`, returns ack, closes.
- \`apps/capture-broker/src/redactor.ts\` — explicit prefix detectors (\`sk-\`, \`xoxb-\`, \`AKIA\`, \`ghp_\`, \`glpat-\`, JWTs, emails) + labeled-context proximity scan for generic 32+ char tokens (spec §5.9 32-char window, whitespace/newlines counted). \`nonce\` and \`v\` (cache-buster) are deliberately excluded from the label set per spec §5.9 carve-outs.
- \`apps/capture-broker/src/bin.ts\` — CLI entrypoint (N3). Production started by systemd \`ExecStart\`. \`--smoke\` flag starts on a temp socket, sends one probe, asserts ok, exits 0 — used in CI.
- \`configs/banner-selectors.yaml\` — curated cookie-banner / GDPR-overlay selector list (~25 selectors, conservative). Owned by browser-automation per spec §5.9.
- \`infra/systemd/willbuy-capture-broker.service\` — systemd unit template (template only; production env file lives in 1Password per CLAUDE.md).

## Review findings addressed (REV REQUEST CHANGES → fixed)

**Blocking:**
- **B1 (socket mode 0660):** `chmodSync(socketPath, 0o660)` called immediately after `server.listen()` resolves in `startBroker()`. This overrides the process umask (systemd `UMask=0007` would produce `0770` without the explicit chmod). Two tests added: inode mode is exactly `0660`; others-bits are `0`. Systemd unit comment updated to reflect that mode is set by the code, not the unit.
- **B2 (proximity window + §6.1 boundary fixtures):** Replaced the immediate-adjacency regex with a proximity scan (`redactLabeledSecrets()`). The separator consumes exactly one `=` or `:` char; the 32-char window is the gap between the separator end and the blob start (whitespace/newlines both count). Added `redactor-boundary.json` fixture suite covering all six spec §6.1 cases: label 31 chars away → redact, label 33 chars away → leave, two labels at 35 and 29 chars → redact (29-char fires), newline in gap → redact, newline+30-spaces = 31 chars → redact, overlapping labels → redact.

**Non-blocking:**
- **N1 (readOneFrame timeout):** `READ_TIMEOUT_MS = 30_000` added. `readOneFrame()` accepts optional `timeoutMs` param (default 30 s). On timeout, socket is destroyed. `BrokerDeps.frameTimeoutMs` allows test injection (300 ms in test suite). Test: broker destroys stalled connection and remains accepting.
- **N2 (worker-side bind-mount):** Not in scope — noted in the systemd unit comment. Will land with capture-worker integration in a follow-up.
- **N3 (dist/bin.js entrypoint):** `src/bin.ts` added. `tsconfig.build.json` + `pnpm build` produces `dist/bin.js`. Smoke test: `dist/bin.js --smoke` exits 0.
- **N4 (.strict() rejects extra fields):** Test added asserting `schema_invalid` on a request with an unknown top-level field. `.strict()` was already in the schema; test confirms it.
- **N5 (screenshot_b64: "" rejected):** Changed `z.string().optional()` → `z.string().min(1).optional()`. Test asserts `schema_invalid` on empty string.

## TDD evidence

Red commit: `212af47` — 8 tests failing (B1/B2/N1/N3/N5 all RED).
Green commit: `8393daf` — 42 tests passing.

## Coverage of the acceptance scenarios

| # | Scenario | Test |
| - | -------- | ---- |
| 1 | Round-trip → storage object exists, row matches | \`server.test.ts > accepts a valid request\` |
| 2 | Oversized a11y_tree → reject, no writes | \`server.test.ts > rejects an oversized a11y_tree\` |
| 3 | Malformed JSON → reject | \`server.test.ts > rejects malformed JSON\` |
| 4 | Multiple messages on one connection → only first accepted | \`server.test.ts > rejects a connection that tries to send two messages\` |
| 5 | Redactor positive fixtures → redacted in stored artifact | \`redactor.test.ts\` (10 cases) + \`server.test.ts > redacts secrets in the stored artifact\` |
| 6 | Redactor false-positive fixtures → NOT redacted | \`redactor.test.ts\` (7 cases) + \`server.test.ts > preserves false-positive strings (40-char asset hash)\` |
| 7 | Banner selectors fixture matches curated list | \`bannerSelectors.test.ts\` (3 cases) + \`server.test.ts > persists banner_selectors_matched\` |
| B1 | Socket inode mode 0660 | \`server.test.ts > socket inode has mode 0660\` + \`socket others-bits are 0\` |
| B2 | §6.1 boundary fixtures (6 cases) | \`redactor.test.ts > §6.1 labeled-context boundary fixtures\` |
| N1 | No-FIN peer connection dropped by timeout | \`server.test.ts > drops stalled connection\` |
| N3 | bin.ts + dist/bin.js + smoke | \`bin.test.ts\` (3 cases) |
| N4 | .strict() rejects unknown fields | \`server.test.ts > rejects a request with an unknown top-level field\` |
| N5 | Empty screenshot_b64 rejected | \`server.test.ts > rejects screenshot_b64 as empty string\` |

## Public-repo audit

- The redactor positive fixtures (\`apps/capture-broker/test/fixtures/redactor-positive.json\`) store \`prefix\` + \`body\` separately so no committed file contains the literal patterns GitHub's secret scanner flags (Slack \`xoxb-…\`, GitLab \`glpat-…\`, etc). The test concatenates them at runtime.
- AWS docs canonical \`AKIAIOSFODNN7EXAMPLE\` is the only documented-fake string we keep verbatim.
- No real secrets, IPs, or VM identifiers in the diff.

## Spec links

- Implements spec §5.13 (Capture Broker design — v0.1 Unix socket transport, schema, byte caps, single-shot, socket mode 0660)
- Implements spec §5.9 (data classification & redaction, labeled-context 32-char proximity rule, §6.1 boundary fixture suite)
- Implements spec §2 #2 (banner DOM-removal selector list — broker forwards; worker removes from DOM, never clicks)

## Test plan

- [x] `pnpm --filter @willbuy/capture-broker run test` — 42 green
- [x] `pnpm --filter @willbuy/capture-broker run build` — dist/bin.js produced
- [x] `node dist/bin.js --smoke` — exits 0
- [x] `pnpm --filter @willbuy/capture-broker run typecheck` — clean
- [x] Public-repo grep — only deliberate fake fixture strings match secret patterns

## Out of scope (follow-ups)

- Capture-worker integration (touches `apps/capture-worker/`; thin follow-up — broker contract is stable)
- Worker-side bind-mount (N2) — READ-ONLY mount of broker.sock into capture container lands with capture-worker integration
- iptables / network namespace egress firewall (S2-9)
- Screenshot OCR scrub (Sprint 3, opt-in per spec §5.17)


apps/capture-broker/src/server.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+  });
+
+  await new Promise<void>((ok, ko) => {
+    server.once('error', ko);
+    server.listen(deps.socketPath, () => {
+      server.removeListener('error', ko);
+      ok();
+    });
+  });
```
NikolayS, line 91: **BLOCKING (B1) — socket mode is not 0660.** `server.listen(socketPath)` creates the socket inode under the process umask. The systemd unit's `UMask=0007` produces mode `0770`, not `0660` as spec §5.13 requires. `RuntimeDirectoryMode=0750` constrains only the directory, not the socket file.

Fix: `await fs.promises.chmod(deps.socketPath, 0o660);` immediately after the `listen` resolves. Or introduce a separate `willbuy-capture-broker.socket` unit with `SocketMode=0660 SocketGroup=willbuy` and switch this to socket-activated `Type=notify`.

This is a direct spec violation and a real-world security regression — any host-local user in the worker container's UID space (or any other container that bind-mounts the same socket) gets full write access in the produced configuration.

apps/capture-broker/src/redactor.ts
```diff
@@ -0,0 +1,163 @@
...
+    // Optional auth scheme + space (Bearer / Basic / Token)
+    String.raw`(?:(?:Bearer|Basic|Token)\s+)?` +
+    // Optional opening quote
+    String.raw`["']?` +
+    // The value: 32+ chars of hex/base64url
+    String.raw`([A-Za-z0-9_\-+/]{32,})`,
+  'gi',
+);
+
+/**
```
NikolayS, line 132: **BLOCKING (B2) — labeled-context detector deviates from spec §5.9 + §6.1.** Spec §6.1 mandates the boundary-fixture suite:

> label 31 chars away → redact, label 33 chars away → don't redact, two labels at 29 and 35 chars → redact, label separated by a newline → redact (newlines counted), multiple overlapping labels

This implementation has no proximity-window scan at all — `RE_LABELED_SECRET` requires the label to **immediately** punctuate the value (`label\s*[:=]\s*…`). Effectively a 0-char window vs. spec's ~32-char window.

Two paths forward:
1. Implement the proximity-window matching the spec, AND add the §6.1 boundary fixtures.
2. Record the deviation in `.samo/spec/willbuy/SPEC.willbuy.amendments.md` (CLAUDE.md: "spec is authoritative; record deviations") justifying the more conservative immediate-adjacency rule, AND add boundary fixtures asserting the chosen behaviour.

Either way the §6.1 boundary fixture suite is a TDD-acceptance contract item that has to land in this PR.

apps/capture-broker/src/framing.ts
```diff
@@ -0,0 +1,123 @@
...
+        if (total > need) {
+          const buf = all();
+          finish({
+            kind: 'ok',
+            payload: buf.slice(HEADER_BYTES, need),
+            trailingBytes: total - need,
+          });
+        }
+      }
+    };
```
NikolayS, line 108: **N1 (non-blocking) — no per-connection timeout.** If a peer sends exactly one valid framed message but never half-closes (no FIN), `total === need` and we hit neither the `total > need` branch nor `onEnd`. The promise never resolves; a buggy/hostile worker can hold connection slots indefinitely.

Add `socket.setTimeout(30_000, () => finish({ kind: 'error', message: 'idle timeout' }))` (or similar) at the top of `readOneFrame`. Spec doesn't mandate a per-connection timeout but operationally this is the right move.

infra/systemd/willbuy-capture-broker.service
```diff
@@ -0,0 +1,47 @@
...
+Description=willbuy capture broker (spec §5.13)
+After=network.target
+
+[Service]
+Type=simple
+User=root
+Group=willbuy
+UMask=0007
+RuntimeDirectory=willbuy
+RuntimeDirectoryMode=0750
```
NikolayS, line 27: **N3 (non-blocking) — `dist/bin.js` does not exist in this PR.** No `src/bin.ts`, no `package.json` `bin` entry, no build step that emits `dist/bin.js`. The unit cannot start as-is. Add a `// TODO follow-up` note here, or land a tiny `bin.ts` (just `import './server.js'; await startBroker({...readEnv()...});`) before this unit is deployed.

apps/capture-broker/src/server.ts
```diff
@@ -0,0 +1,199 @@
...
+  } catch (e) {
+    sendError('db_failed', e instanceof Error ? e.message.slice(0, 200) : undefined);
+    return;
+  }
+
+  const ackOk: BrokerAck = screenshotKey
+    ? { ok: true, capture_id: captureId, a11y_object_key: a11yKey, screenshot_object_key: screenshotKey }
+    : { ok: true, capture_id: captureId, a11y_object_key: a11yKey };
+  sendAck(ackOk);
+}
```
NikolayS, line 226: **N4 (non-blocking) — add a test for `.strict()` schema.** The schema correctly rejects extra fields (`schema.ts` uses `.strict()`), but `test/server.test.ts` doesn't assert that contract. One small test: send `{...validRequest, extra_unknown_field: 1}` and expect `schema_invalid`. Locks the contract against accidental loosening to `.passthrough()`.

apps/capture-broker/src/server.ts
```diff
@@ -0,0 +1,199 @@
...
+    screenshotBytes = decodedBase64Bytes(req.screenshot_b64);
+    if (screenshotBytes === null) {
+      sendError('schema_invalid', 'screenshot_b64 is not valid base64');
+      return;
+    }
+    if (screenshotBytes > BYTE_CAPS.SCREENSHOT_BYTES) {
+      sendError('screenshot_too_big', `decoded ${screenshotBytes} > cap ${BYTE_CAPS.SCREENSHOT_BYTES}`);
+      return;
+    }
+  }
```
NikolayS, line 176: **N5 (non-blocking) — `screenshot_b64: ""` writes a 0-byte object.** Empty string passes schema (`z.string().optional()`) and `decodedBase64Bytes('')` returns 0, then the worker happily writes `screenshot.png` with 0 bytes to storage. v0.1 keeps screenshots disabled so this is latent, but worth tightening: `screenshot_b64: z.string().min(1).optional()` or refine.

NikolayS (commented):
## FINAL VERDICT — REQUEST CHANGES

Tested the PR locally end-to-end:

```
$ pnpm --filter @willbuy/capture-broker run test
 ✓ test/redactor.test.ts (17 tests) 3ms
 ✓ test/bannerSelectors.test.ts (3 tests) 5ms
 ✓ test/server.test.ts (8 tests) 550ms
 Test Files  3 passed (3)
      Tests  28 passed (28)
```

Red-state verified at `7a01471` (27 fail / 1 pass) → green at `d18f07e` (28 pass). TDD discipline is clean. The architecture (length-prefixed framing, in-memory test doubles for storage + DB, schema-then-byte-cap split) is well-executed.

But two findings are blocking.

### Blocking findings (2)

**B1 — Socket inode mode is NOT 0660 (security).** Spec §5.13 explicitly demands `mode 0660`. The systemd unit's `UMask=0007` would yield `0770` for the socket inode, and there's nothing in `startBroker()` that calls `chmodSync(socketPath, 0o660)` after `server.listen()`. `RuntimeDirectoryMode=0750` controls the directory, not the socket file. Without an explicit `chmod` call (or a systemd socket-activation unit with `SocketMode=0660`), the broker socket goes into production with the wrong permission bits — directly violating the spec. The PR description claim "mode 0660 set by the systemd unit" is incorrect.

Fix: add `chmodSync(deps.socketPath, 0o660)` immediately after `listen` resolves, OR introduce a `willbuy-capture-broker.socket` unit with `SocketMode=0660` + `SocketGroup=willbuy` and switch the service to `Type=notify` socket activation.

**B2 — Spec §6.1 boundary fixtures missing; redactor deviates from labeled-context proximity rule.** Spec §6.1 requires the boundary-fixture suite for `redactCapture`:

> label 31 chars away → redact, label 33 chars away → don't redact, two labels at 29 and 35 chars → redact, label separated by a newline → redact (newlines counted), multiple overlapping labels

None of these are in `test/fixtures/redactor-*.json`. Worse, the regex `RE_LABELED_SECRET` only matches when the label *immediately* punctuates the value (`label\s*[:=]\s*(Bearer\s+)?["']?VALUE`). The implementation has no proximity-window scan — effectively a 0-char window vs. the spec's 32-char window. So even if the boundary fixtures were added, half of them would fail.

Two acceptable resolutions:
1. Implement the proximity-window matching the spec describes and add the §6.1 boundary fixtures (preferred), OR
2. Record the deviation in `.samo/spec/willbuy/SPEC.willbuy.amendments.md` (per CLAUDE.md "spec is authoritative") explaining why the conservative immediate-adjacency rule is being adopted instead, and still add the boundary fixtures asserting the chosen behavior.

Either way the §6.1 fixture suite has to land before merge — that's a TDD-acceptance contract item.

### Non-blocking findings (5)

**N1 — `readOneFrame` has no per-connection timeout.** A peer that sends one valid framed message but does NOT half-close (no FIN) will hold the connection forever — neither `total > need` (because `total === need` exactly) nor `onEnd` ever fires. A buggy or hostile worker could exhaust connection slots. Add a 30 s socket timeout + cleanup. Not in spec but warranted.

**N2 — Worker-side bind-mount config not in PR.** Spec §5.13 requires the socket be "bind-mounted READ-ONLY into the container at a fixed in-container path." Acknowledged as out-of-scope ("thin follow-up") but worth a tracking issue so the read-only bind-mount + mode-0660 group-membership end-to-end smoke test lands before this is wired into production.

**N3 — `dist/bin.js` referenced but absent.** `infra/systemd/willbuy-capture-broker.service` `ExecStart`s `/opt/willbuy/capture-broker/dist/bin.js`, but the PR adds no `bin.ts` / `bin.js` entrypoint. The unit cannot start as-is. OK if explicitly deferred; please add a follow-up issue and a TODO marker in the unit.

**N4 — No test for `.strict()` schema rejecting extra fields.** Schema correctly uses `.strict()` so unknown keys raise `schema_invalid`; add one test asserting that to lock the contract (e.g. `{...validRequest, __proto__: {x:1}}` or extra unknown key). Trivial follow-on.

**N5 — `screenshot_b64: ""` writes a 0-byte object.** Empty string passes both schema (`z.string().optional()`) and `decodedBase64Bytes("")` (returns 0). Tighten with `.min(1)` or `.refine` if screenshots ever go live in v0.1; today studies are screenshots-off so latent.

### Rubric summary

| Persona | Outcome |
| --- | --- |
| Security | **B1 socket mode**; otherwise systemd hardening, ownership scheme, and storage/DB injection are clean. |
| Bug Hunter | N1 hang risk; oversize / single-shot / malformed-JSON paths all correctly reject; banner-list is DOM-removal not click. |
| Test Analyzer | TDD red→green verified locally (27→28). 7 acceptance scenarios all covered. **B2 §6.1 boundary fixtures missing.** |
| Guidelines | Storage + DB stubbed (no real Supabase / Postgres in CI) ✓. systemd unit follows hardening pattern ✓. N3 missing entrypoint ⚠. |
| Docs | Spec §5.13 / §5.9 / §2 #2 linked; `Closes #32`; public-repo audit grep clean (no real secrets in diff). ✓ |

SOC2 exception applied per CLAUDE.md.

Address B1 + B2 and re-request review. Do not merge.

NikolayS:
Manager: REV blocked; sending back to fix agent. Blocking items B1 (socket mode 0660 — `chmod` after `listen`) and B2 (redactor proximity model + spec §6.1 boundary fixtures) must land before re-review. Non-blocking N1-N5 also worth addressing in the same PR.

NikolayS:
## Fixes pushed — all 7 findings addressed

**Blocking (B1 + B2):**

- **B1 (socket mode 0660):** `chmodSync(socketPath, 0o660)` called immediately after `server.listen()` resolves. Tests: `socket inode has mode 0660` + `socket others-bits are 0` (both green). Systemd unit comment corrected — mode is set by code, not the unit.

- **B2 (proximity window + §6.1 fixtures):** Replaced the immediate-adjacency regex with a character-distance proximity scan (`redactLabeledSecrets()`). Separator = one `=` or `:` char only; the 32-char window is the gap between separator and blob start (whitespace + newlines counted). Added `test/fixtures/redactor-boundary.json` with 6 cases exactly matching spec §6.1: label 31 chars → redact, label 33 chars → leave, two labels (35 + 29 chars) → redact (29-char fires), newline in gap → redact, newline+30-spaces = 31 → redact, overlapping labels → redact.

**Non-blocking (N1–N5):**

- **N1:** `READ_TIMEOUT_MS = 30_000` in `framing.ts`. `readOneFrame()` takes optional `timeoutMs`. `BrokerDeps.frameTimeoutMs` for test injection (300 ms). Test verifies broker drops stalled connection and remains accepting.
- **N2:** Not in scope — noted in systemd unit comment as follow-up with capture-worker integration.
- **N3:** `src/bin.ts` added (production + `--smoke` mode). `tsconfig.build.json` wired; `pnpm build` produces `dist/bin.js`. Smoke test: `dist/bin.js --smoke` exits 0.
- **N4:** Test added asserting `schema_invalid` on unknown top-level field (`.strict()` was already there; test locks the contract).
- **N5:** `z.string().min(1).optional()` on `screenshot_b64`. Test asserts `schema_invalid` on empty string.

**TDD discipline:**
- Red commit: `212af47` — 8 tests failing (B1/B2/N1/N3/N5 RED, existing 28 still green → 34 total, 8 fail)
- Green commit: `8393daf` — 42 tests passing (28 original + 14 new)

Public-repo audit: no real secrets, IPs, or VM identifiers in diff.

NikolayS:
@NikolayS All B1/B2 blocking + N1-N5 non-blocking findings addressed. 42 tests green (TDD red→green verified). Please re-review.

NikolayS (commented):
## VERDICT: REQUEST CHANGES

Re-review of the B1/B2/N1–N5 fix batch (commits `212af47` RED + `8393daf` GREEN).

---

### TDD discipline — PASS

`212af47` (RED) touches only test/fixture files: `test/bin.test.ts`, `test/fixtures/redactor-boundary.json`, `test/helpers.ts`, `test/redactor.test.ts`, `test/server.test.ts`. Zero implementation files. `8393daf` (GREEN) touches only implementation + a few fixture/test files that needed updating. Discipline is clean.

---

### B1 — Socket mode 0660 — FIXED ✓

`apps/capture-broker/src/server.ts:88` — `chmodSync(deps.socketPath, SOCKET_MODE)` is called inside the `server.listen()` callback, which is the normal-completion path only (the `once('error', ko)` listener is removed before calling `ok()`). The chmod is not on an error path. `SOCKET_MODE = 0o660` is exported at line 13.

Test coverage in `test/server.test.ts`:
- Line 227–232: `statSync(socketPath).mode & 0o777 === 0o660` — asserts the on-disk inode bits via `fs.statSync`. ✓
- Line 237–241: `st.mode & 0o007 === 0` — asserts others-bits are clear (world-no-connect). ✓

Note: the test asserting that a non-`willbuy`-group process is rejected by the OS at `connect()` time is not present (that would require spawning a child under a different UID, which is impractical in a local test). The mode-bit assertion is the correct and standard substitute. Acceptable.

**B1: resolved.**

---

### B2 — Redactor proximity model + §6.1 boundary fixtures — FIXED ✓

`apps/capture-broker/src/redactor.ts` — `redactLabeledSecrets()` (lines 153–218) implements a proximity scan:

- Label anchor regex (`RE_LABEL_ANCHOR`, lines 118–127) matches a label keyword + exactly one `=` or `:` separator. Lookbehind `(?<=[^A-Za-z0-9_])` prevents compound words like `csrf_token` from matching via the bare `token` keyword.
- `LABEL_PROXIMITY_CHARS = 32` (line 110). Window is `[sepEnd, sepEnd + 32)`. Gap chars (whitespace, newlines) all count toward the distance. ✓
- Blob search starts from `sepEnd` in the full string; blob accepted only if `blobMatch.index < windowEnd` (line 182). Blob may extend beyond the window; only the start position is checked. ✓

`nonce` is deliberately absent from `LABEL_KEYWORDS` (line 89–105), so `nonce="<64-char-hex>"` is not triggered. `v` is also absent, so `?v=<32-char-hex>` cache-busters are safe. ✓

Boundary fixtures (`test/fixtures/redactor-boundary.json`) — all 6 cases verified:

| Case | Gap | Expected | Math |
|---|---|---|---|
| 31-char gap | 31 spaces | redact | bodyPos 37 < windowEnd 38 ✓ |
| 33-char gap | 33 spaces | leave | bodyPos 39 ≥ windowEnd 38 ✓ |
| two labels (29 + 35) | varies | redact | `token=` fires at gap 29 ✓ |
| newline in gap | `\n` (1 char) | redact | gap 1 ≤ 32 ✓ |
| newline + 30 spaces | 31 chars | redact | gap 31 ≤ 32 ✓ |
| overlapping labels | 0 | redact | `secret=token=` fires ✓ |

False-positive carve-outs (`test/fixtures/redactor-false-positive.json`) — all 5 spec §5.9 classes present: 40-char asset-URL hash, 64-char CSP nonce, 64-char SKU, commit SHA (40-char), cache-busted image URL. Plus 2 extra (version string hash, CSS URL hash). All 7 cases have no label context adjacent to the blob, so no detector fires. ✓

**B2: resolved.**

---

### N1 — Timeout — FIXED ✓

`apps/capture-broker/src/framing.ts:31` — `READ_TIMEOUT_MS = 30_000` exported.
`BrokerDeps.frameTimeoutMs` injectable (server.ts:44). Tests inject `frameTimeoutMs: 300`.
`test/server.test.ts:248–268` — `sendOnceNoEnd` sends a framed message without half-closing; after 300 ms the broker destroys the socket; helper throws; test asserts `gotError === true` then confirms the server still accepts a follow-up request. ✓

**N1: resolved.**

---

### N3 — Entrypoint — PARTIALLY FIXED — NEW BLOCKING ISSUE

`apps/capture-broker/src/bin.ts` exists. `apps/capture-broker/tsconfig.build.json` exists.
`test/bin.test.ts` — smoke test spawns `node dist/bin.js --smoke` and asserts exit 0.

**The smoke test passes when `pnpm build` is run first (locally confirmed: 42/42 green after build). However, CI never runs `pnpm build` before `pnpm test`.**

`.github/workflows/ci.yml` steps (lines 44–51):
```
- name: Typecheck
  run: pnpm typecheck
- name: Lint
  run: pnpm lint
- name: Test
  run: pnpm test
```

There is no `Build` step. The root `tsconfig.json` `references` array does not include `capture-broker`, so `tsc -b` at the Typecheck step does not emit `dist/bin.js`. The `pnpm test` step invokes each package's own `vitest run` (no pre-build). Result: `dist/bin.js` does not exist when `bin.test.ts` runs in CI, causing two test failures:

- `capture-broker entrypoint (N3) > dist/bin.js exists after pnpm build` — FAILS
- `capture-broker entrypoint (N3) > dist/bin.js --smoke exits 0` — FAILS

**Fix required:** add a `Build` step to `.github/workflows/ci.yml` between Typecheck and Lint (or between Lint and Test):
```yaml
- name: Build
  run: pnpm build
```

---

### BLOCKING — Lint failure: unused variable `window` in redactor.ts

**`apps/capture-broker/src/redactor.ts:173`**

```typescript
const window = input.slice(sepEnd, windowEnd);   // ← assigned, never read
```

This variable is computed but never referenced. The actual boundary check at line 182 uses `windowEnd` directly:
```typescript
if (blobMatch.index >= windowEnd) continue;
```

ESLint reports: `'window' is assigned a value but never used  @typescript-eslint/no-unused-vars`

This is the root cause of the current CI red state (`build` job → `Lint` step → exit 1). Confirmed from `gh run view 24932833631 --log-failed`.

**Fix:** remove line 173 entirely (the slice is not needed for the proximity check) or prefix with `_` if it was retained for debugging readability:
```typescript
// Remove this line:
const window = input.slice(sepEnd, windowEnd);
```

---

### N4 — Zod `.strict()` — FIXED ✓

`apps/capture-broker/src/schema.ts:33` — `.strict()` is on the `CaptureRequest` schema. Test at `server.test.ts:271–277` sends an unknown field and expects `schema_invalid`. ✓

---

### N5 — Empty screenshot string — FIXED ✓

`apps/capture-broker/src/schema.ts:26` — `screenshot_b64: z.string().min(1).optional()`. Test at `server.test.ts:281–287`. ✓

---

### Summary of findings

| # | Severity | File:line | Finding |
|---|---|---|---|
| F1 | **BLOCKING** | `apps/capture-broker/src/redactor.ts:173` | Unused variable `window` → ESLint `no-unused-vars` → CI lint step fails |
| F2 | **BLOCKING** | `.github/workflows/ci.yml:44–51` | No `pnpm build` step before `pnpm test` → N3 smoke tests always fail in CI once lint is fixed |
| — | resolved | — | B1 socket mode 0660 ✓ |
| — | resolved | — | B2 proximity redactor + §6.1 fixtures ✓ |
| — | resolved | — | N1 timeout ✓ |
| — | resolved | — | N4 `.strict()` ✓ |
| — | resolved | — | N5 empty screenshot ✓ |

Two new blocking issues introduced by the fix batch. Both are one-liners to fix. Once `redactor.ts:173` is removed and a `Build` step is added to `ci.yml`, all findings are resolved and this is approvable.

NikolayS:
## VERDICT: REQUEST CHANGES

Re-review of the B1/B2/N1–N5 fix batch (commits `212af47` RED + `8393daf` GREEN).

---

### TDD discipline — PASS

`212af47` (RED) touches only test/fixture files: `test/bin.test.ts`, `test/fixtures/redactor-boundary.json`, `test/helpers.ts`, `test/redactor.test.ts`, `test/server.test.ts`. Zero implementation files. `8393daf` (GREEN) touches only implementation + a few fixture/test files that needed updating. Discipline is clean.

---

### B1 — Socket mode 0660 — FIXED ✓

`apps/capture-broker/src/server.ts:88` — `chmodSync(deps.socketPath, SOCKET_MODE)` is called inside the `server.listen()` callback, which is the normal-completion path only (the `once('error', ko)` listener is removed before calling `ok()`). The chmod is not on an error path. `SOCKET_MODE = 0o660` is exported at line 13.

Test coverage in `test/server.test.ts`:
- Line 227–232: `statSync(socketPath).mode & 0o777 === 0o660` — asserts the on-disk inode bits via `fs.statSync`. ✓
- Line 237–241: `st.mode & 0o007 === 0` — asserts others-bits are clear (world-no-connect). ✓

Note: the test asserting that a non-`willbuy`-group process is rejected by the OS at `connect()` time is not present (that would require spawning a child under a different UID, which is impractical in a local test). The mode-bit assertion is the correct and standard substitute. Acceptable.

**B1: resolved.**

---

### B2 — Redactor proximity model + §6.1 boundary fixtures — FIXED ✓

`apps/capture-broker/src/redactor.ts` — `redactLabeledSecrets()` (lines 153–218) implements a proximity scan:

- Label anchor regex (`RE_LABEL_ANCHOR`, lines 118–127) matches a label keyword + exactly one `=` or `:` separator. Lookbehind `(?<=[^A-Za-z0-9_])` prevents compound words like `csrf_token` from matching via the bare `token` keyword.
- `LABEL_PROXIMITY_CHARS = 32` (line 110). Window is `[sepEnd, sepEnd + 32)`. Gap chars (whitespace, newlines) all count toward the distance. ✓
- Blob search starts from `sepEnd` in the full string; blob accepted only if `blobMatch.index < windowEnd` (line 182). Blob may extend beyond the window; only the start position is checked. ✓

`nonce` is deliberately absent from `LABEL_KEYWORDS` (lines 89–105), so `nonce="<64-char-hex>"` is not triggered. `v` is also absent, so `?v=<32-char-hex>` cache-busters are safe. ✓

Boundary fixtures (`test/fixtures/redactor-boundary.json`) — all 6 cases verified:

| Case | Gap | Expected | Math |
|---|---|---|---|
| 31-char gap | 31 spaces | redact | bodyPos 37 < windowEnd 38 ✓ |
| 33-char gap | 33 spaces | leave | bodyPos 39 ≥ windowEnd 38 ✓ |
| two labels (29 + 35) | varies | redact | `token=` fires at gap 29 ✓ |
| newline in gap | `\n` (1 char) | redact | gap 1 ≤ 32 ✓ |
| newline + 30 spaces | 31 chars | redact | gap 31 ≤ 32 ✓ |
| overlapping labels | 0 | redact | `secret=token=` fires ✓ |

False-positive carve-outs (`test/fixtures/redactor-false-positive.json`) — all 5 spec §5.9 classes present: 40-char asset-URL hash, 64-char CSP nonce, 64-char SKU, commit SHA (40-char), cache-busted image URL. Plus 2 extra (version string hash, CSS URL hash). All 7 cases have no label context adjacent to the blob, so no detector fires. ✓

**B2: resolved.**

---

### N1 — Timeout — FIXED ✓

`apps/capture-broker/src/framing.ts:31` — `READ_TIMEOUT_MS = 30_000` exported.
`BrokerDeps.frameTimeoutMs` injectable (server.ts:44). Tests inject `frameTimeoutMs: 300`.
`test/server.test.ts:248–268` — `sendOnceNoEnd` sends a framed message without half-closing; after 300 ms the broker destroys the socket; helper throws; test asserts `gotError === true` then confirms the server still accepts a follow-up request. ✓

**N1: resolved.**

---

### N3 — Entrypoint — PARTIALLY FIXED — NEW BLOCKING ISSUE

`apps/capture-broker/src/bin.ts` exists. `apps/capture-broker/tsconfig.build.json` exists.
`test/bin.test.ts` — smoke test spawns `node dist/bin.js --smoke` and asserts exit 0.

**The smoke test passes when `pnpm build` is run first (locally confirmed: 42/42 green after build). However, CI never runs `pnpm build` before `pnpm test`.**

`.github/workflows/ci.yml` steps (lines 44–51):
```yaml
- name: Typecheck
  run: pnpm typecheck
- name: Lint
  run: pnpm lint
- name: Test
  run: pnpm test
```

There is no `Build` step. The root `tsconfig.json` `references` array does not include `capture-broker`, so `tsc -b` at the Typecheck step does not emit `dist/bin.js`. The `pnpm test` step invokes each package's own `vitest run` (no pre-build). Result: `dist/bin.js` does not exist when `bin.test.ts` runs in CI.

**Fix required:** add a `Build` step to `.github/workflows/ci.yml`:
```yaml
- name: Build
  run: pnpm build
```
(Between Lint and Test, or between Typecheck and Lint.)

---

### BLOCKING — Lint failure: unused variable `window` in redactor.ts

**`apps/capture-broker/src/redactor.ts:173`**

```typescript
const window = input.slice(sepEnd, windowEnd);   // ← assigned, never read
```

This variable is computed but never referenced. The actual boundary check at line 182 uses `windowEnd` directly:
```typescript
if (blobMatch.index >= windowEnd) continue;
```

ESLint reports: `'window' is assigned a value but never used  @typescript-eslint/no-unused-vars`

This is the root cause of the current CI red state (`build` job → `Lint` step → exit 1). Confirmed from CI run logs.

**Fix:** remove line 173 entirely (the slice is not needed for the proximity check).

---

### N4 — Zod `.strict()` — FIXED ✓

`apps/capture-broker/src/schema.ts:33` — `.strict()` is on the `CaptureRequest` schema. Test at `server.test.ts:271–277` sends an unknown field and expects `schema_invalid`. ✓

---

### N5 — Empty screenshot string — FIXED ✓

`apps/capture-broker/src/schema.ts:26` — `screenshot_b64: z.string().min(1).optional()`. Test at `server.test.ts:281–287`. ✓

---

### Summary

| # | Severity | Location | Finding |
|---|---|---|---|
| F1 | **blocking** | `apps/capture-broker/src/redactor.ts:173` | Unused `window` variable → ESLint `no-unused-vars` → CI lint fails |
| F2 | **blocking** | `.github/workflows/ci.yml:44–51` | No `pnpm build` step → `dist/bin.js` never created → N3 smoke tests fail in CI |
| B1 | resolved | `server.ts:88` + `server.test.ts:227–241` | chmod + mode-bit assertions ✓ |
| B2 | resolved | `redactor.ts:153–218` + boundary fixtures | Proximity scan + §6.1 cases ✓ |
| N1 | resolved | `framing.ts:31` + `server.test.ts:248–268` | Timeout fires + server survives ✓ |
| N4 | resolved | `schema.ts:33` | `.strict()` present ✓ |
| N5 | resolved | `schema.ts:26` | `min(1)` present ✓ |

Two new blocking issues introduced by the fix batch. Both are one-liners. Once `redactor.ts:173` is removed and a `Build` step is added to `ci.yml`, all findings are resolved and this is approvable.

NikolayS:
## F1 + F2 fixes — commit 3069d9e

Two new blockers introduced by the previous fix are now resolved.

---

### F1 — lint failure (`redactor.ts:173`) ✅

**Root cause:** `const window = input.slice(sepEnd, windowEnd)` was assigned but never read — the boundary check at line 182 uses `windowEnd` directly. ESLint `@typescript-eslint/no-unused-vars` fired.

**Fix:** Deleted the unused `const window = ...` line in `apps/capture-broker/src/redactor.ts`.

**Verified:** `pnpm lint` → clean (0 errors). All 42 capture-broker tests still pass.

---

### F2 — smoke test broken in CI (`bin.test.ts` / CI workflow) ✅

**Root cause:** Two issues:
1. CI had no `pnpm build` step before `pnpm test`, so `dist/bin.js` never existed when the N3 smoke test (`bin.test.ts`) ran.
2. Root `tsconfig.json` references didn't include `capture-broker`, so `tsc -b` at the root wouldn't track its build artefact.

**Fix (Option A):**
- Added `"composite": true` to `apps/capture-broker/tsconfig.build.json` (required for TypeScript project references).
- Added `{ "path": "./apps/capture-broker/tsconfig.build.json" }` to the root `tsconfig.json` references array (pointing to the build config that emits to `dist/`, not the typecheck-only `tsconfig.json`).
- Added `pnpm build` step in `.github/workflows/ci.yml` between the Lint and Test steps.

**Verified locally:** `pnpm build` → all packages (api, capture-broker, web) build cleanly. `pnpm test` → all 42 capture-broker tests pass including the 3 `bin.test.ts` smoke tests. `tsc -b --dry` correctly picks up `apps/capture-broker/tsconfig.build.json` as a project reference.

---

@NikolayS — ready for re-review. CI run will confirm green on push.

NikolayS:
## VERDICT: APPROVE

**rev-pr41-v3** — re-review after F1+F2 CI fix. No new findings. All previously-cleared items confirmed intact.

---

### CI Status

All three jobs **GREEN** on `d50cda7` (latest HEAD):
- `build` — pass (2 m 6 s)
- `aggregator` — pass (3 m 18 s)
- `egress-integration` — pass (34 s)

---

### F1 — Unused `window` var (CLOSED)

`apps/capture-broker/src/redactor.ts` line 173: `const window = input.slice(sepEnd, windowEnd)` is **gone**. The boundary guard at line 181 uses `windowEnd` directly as it should. No `// eslint-disable-next-line` waivers added anywhere in `src/`. Fix is surgical: 1-line deletion, correct logic preserved.

### F2 — Build wiring (CLOSED)

Three-part fix all verified:

1. `.github/workflows/ci.yml` — `pnpm build` step runs after `Lint` and before `Test` (lines 50–54). Order is correct.
2. Root `tsconfig.json` — `references` array now includes `{ "path": "./apps/capture-broker/tsconfig.build.json" }`.
3. `apps/capture-broker/tsconfig.build.json` — `composite: true`, `noEmit: false`, `outDir: "dist"`. Emits correctly.
4. `apps/capture-broker/test/bin.test.ts` — three tests: `src/bin.ts exists`, `dist/bin.js exists after pnpm build`, and `dist/bin.js --smoke exits 0` (exercises the full runtime path via `node dist/bin.js`, not tsx). Meaningful smoke test confirmed.

### Merge from main (CLEAN)

Merge commit `d50cda7` (Merge: `55d7a57` ← `0d0cf04`) brings in:
- Migrations 0001..0011 (`infra/migrations/`) — all 12 files present.
- Egress wiring (`apps/capture-worker/src/run-with-netns.ts`, infra shell suites).
- Aggregator / report-viz / llm-adapter transport-retry.
- SPEC amendments + precedence callout.

No revert-merges or "ours"-strategy artifacts. No feat/32-broker functionality was dropped.

### Previously-cleared findings — all confirmed present

| ID | Check | File | Status |
|----|-------|------|--------|
| B1 | `chmodSync(socketPath, 0o660)` after listen() | `server.ts` line 88 | ✓ |
| B2 | Proximity 32-char window (`LABEL_PROXIMITY_CHARS`) | `redactor.ts` | ✓ |
| N1 | `READ_TIMEOUT_MS = 30_000` + injectable `frameTimeoutMs` | `framing.ts` line 31, `server.ts` line 44 | ✓ |
| N3 | `--smoke` flag in `bin.ts`; `dist/bin.js --smoke` test | `bin.ts` line 23, `test/bin.test.ts` | ✓ |
| N4 | `.strict()` on `CaptureRequest` | `schema.ts` line 33 | ✓ |
| N5 | `screenshot_b64: z.string().min(1).optional()` | `schema.ts` line 26 | ✓ |

### Acknowledged (not findings)

- `172495a ci: trigger CI run for F1+F2 fixes` + `55d7a57 Revert "ci: trigger CI run for F1+F2 fixes"` — empty commit + revert pair to force CI on a stuck push. Harmless.

---

**Finding count: 0.** Ready to merge at manager's discretion.

PR #56, opened 2026-04-25
fix(llm-adapter): extend FORBIDDEN to 9 identifiers + amendment A6 (#20)

NikolayS:
## Root cause

Sprint 1 alignment audit found `context_id` and `assistant_id` missing from §2 #12 enforcement. The `FORBIDDEN` Set in `packages/llm-adapter/eslint-rule.js` had only 7 identifiers; spec §2 #12 authoritatively lists 9.

## Changes

**BD-1 — Code fix (TDD red→green):**
- `packages/llm-adapter/eslint-rule.js`: added `context_id` and `assistant_id` to the `FORBIDDEN` Set (now 9 entries); updated doc-comment and lint message to reference §2 #12 (not #15)
- `packages/llm-adapter/lint-fixtures/forbidden-context-id.ts`: RED fixture — `context_id` as object-literal property key
- `packages/llm-adapter/lint-fixtures/forbidden-assistant-id.ts`: RED fixture — `assistant_id` as TS interface property signature
- `packages/llm-adapter/test/lint-rule.test.ts`: two new test cases for the above fixtures
- `CLAUDE.md` line 45: updated grep-pattern and spec ref to include all 9 identifiers (§2 #12)

**A6 — Spec amendment:**
- `.samo/spec/willbuy/SPEC.willbuy.amendments.md`: appended A6 recording that `packages/adapters/` was renamed to `packages/llm-adapter/` and the forbidden list lives inline in `eslint-rule.js` instead of `forbidden-keys.ts`

## Test evidence — lint-rule (all 9 identifiers, 8/8 passing)

```
 RUN  v2.1.9 /Users/nik/github/willbuy-wt-20-forbidden-keys

 ✓ packages/llm-adapter/test/lint-rule.test.ts (8 tests) 60333ms
   ✓ willbuy/no-reserved-llm-identifiers — forbidden fixtures > flags a top-level `const session_id = …` declaration 9125ms
   ✓ willbuy/no-reserved-llm-identifiers — forbidden fixtures > flags `conversation_id` in a TS interface property 5687ms
   ✓ willbuy/no-reserved-llm-identifiers — forbidden fixtures > flags `thread_id` as an object-literal property 9218ms
   ✓ willbuy/no-reserved-llm-identifiers — forbidden fixtures > flags `previous_response_id` as a function parameter 5820ms
   ✓ willbuy/no-reserved-llm-identifiers — forbidden fixtures > flags `run_id` inside an object destructure 8506ms
[...]

PR #54, opened 2026-04-25
feat(api): per-backstory lease + single-writer finalize (#29)

NikolayS:
## Summary

- Implements spec §2 #12 (per-backstory lease) and §5.11 (single-writer aggregator lock with `SELECT … FOR UPDATE SKIP LOCKED`) per the v0.5 amendment
- `apps/api/src/leases/backstory-lease.ts`: `acquireLease`, `releaseLease`, `extendLease`, `acquireLeaseWithBackoff` — uses `FOR UPDATE` transaction for correct reclaim-or-fail semantics with expired leases; `holder_visit_id` is a typed FK per migration B4 (0005_visits.sql)
- `apps/api/src/finalize/aggregator-lock.ts`: `acquireFinalizeLock` (`FOR UPDATE SKIP LOCKED`), `commitReport` (writes `reports` row + sets `studies.status='ready'`), `failStudy`, `recordLateArrival` (idempotent)
- 13 real-DB integration tests (no mocks) covering all 5 acceptance criteria

## SKIP LOCKED choice — spec reference

Per spec §5.11 (v0.5 amendment, 2026-04-24): `SKIP LOCKED` frees the connection-pool slot immediately when contention is detected. `NOWAIT` was considered and rejected for noisier log semantics. Correctness of "exactly one report" is guaranteed by `UNIQUE(reports.study_id)`, not by the lock.

## Concurrency bench (AC#3) — 10 concurrent aggregators

```
[bench] 10 concurrent aggregators completed in 9 ms; 1 winner, 9 losers (zero-row SKIP LOCKED)
```

Exactly 1 winner commits `ready`; 9 losers receive zero rows from `FOR UPDATE SKIP LOCKED` and return within the same connection-pool round-trip (~1 ms each). No aggregator blocks for the aggregation duration. `UNIQUE(reports.study_id)` is the backstop correctness guarantee.

## Paired A/B isolation

While side A's lease is held (valid `lease_until`), side B's `acquireLease` returns `{ ok: false, reason: 'held' }` immediately — verified by AC#1 tests. The two variants can never be in-flight simultaneously for the same backstory.

## Spec sections satisfied

- §2 #12 (per-backstory lease row, heartbeat, 120 s TTL, release on terminal commit)
- §2 #18 (paired-A/B isolation via lease)
- §5.11 (single-writer finalize, SKIP LOCKED, `late_arrivals`, 3-min timeout context)
- v0.5 amendment (SKIP LOCKED vs NOWAIT rationale — referenced in code comments)

[...]

PR #60, opened 2026-04-25
test(infra): fix migrations testcontainer startup race (#57)

### LINKED ISSUE #57: test(infra): flaky migrations.test.ts — Postgres testcontainer startup race
**Symptom (recurring on multiple PRs):**
\`\`\`
psql: error: connection to server at "127.0.0.1", port 32897 failed: FATAL: the database system is starting up
psql: error: connection to server at "127.0.0.1", port 33563 failed: server closed the connection unexpectedly
\`\`\`

The `tests/migrations.test.ts` and `tests/migrations.schema.test.ts` suites use Testcontainers to spin up a Postgres container per test file. The test executes `psql` against the container before Postgres has finished initial startup, OR Postgres dies mid-test under runner resource pressure.

**Observed on:**
- PR #54 (per-backstory lease) — initially blamed on GH-services + testcontainers collision; engineer fixed by removing the service, but the flake recurred elsewhere.
- PR #56 (forbidden-keys) — the change is ESLint-only, the failure is irrelevant to the PR's diff.

**Likely root causes:**
1. Test calls `psql` before container's healthcheck reports ready. Need to use Testcontainers' `waitStrategy.forLogMessage` or `waitForPort` correctly.
2. Postgres OOM-killed by limited GH Actions runner memory. Could pin a smaller `postgres:16-alpine` image OR set `shared_buffers=128MB` to reduce memory footprint.
3. Test parallelism is too aggressive — multiple Postgres testcontainers + the build job all eating memory simultaneously.

**Fix candidates:**
- Add `Wait.forLogMessage('database system is ready to accept connections')` (or its testcontainers-node equivalent) to the container `start()` config in `tests/migrations.test.ts`.
- Use `vitest.config.ts`'s `test.poolOptions.threads.singleThread = true` for the migration suite to avoid parallel Postgres instances.
- Pin a lower-resource Postgres image.

Pick whichever combination passes 10 consecutive CI runs without a re-run.

**Out of scope for any feature PR.** Should be its own small infra PR. Until fixed, treat the failure as a known flake on PRs that don't touch `tests/migrations.test.ts` or related schema files.

NikolayS:
## Root cause

`pg_isready` returns exit 0 as soon as the postmaster socket exists — before `initdb` finishes and the DB accepts SQL. This caused the `psql: FATAL: the database system is starting up` errors on PRs #54 and #56. Compounding this, when the full vitest suite runs, two concurrent Postgres containers (`migrations.test.ts` + `migrations.schema.test.ts`) consumed enough memory to trigger OOM-kills on the runner.

## Fixes (all three applied)

**1. Wait strategy** — `startPostgres()` now polls `docker logs` for the literal line `"database system is ready to accept connections"` before issuing `pg_isready`. This is the canonical signal that `initdb` has completed and the postmaster is fully up. Deadline extended from 30 s → 60 s for slow runners.

**2. Image hardening** — added to `docker run` args in both test files:
- `POSTGRES_INITDB_ARGS=--auth-host=trust` — skips md5 auth handshake during initdb, faster startup
- `POSTGRES_SHARED_BUFFERS=128MB` + `--shm-size=256m` — smaller memory footprint, prevents OOM-kills under CI memory pressure

**3. Vitest serialization** — `vitest.config.ts` now sets `poolOptions.threads.singleThread: true, maxThreads: 1` so the two migration suites never spin up concurrent Postgres containers in the same vitest run.

## Local test run (proof)

```
$ bun test tests/migrations.test.ts
bun test v1.3.5 (1e86cebd)

 3 pass
 0 fail
 15 expect() calls
Ran 3 tests across 1 file. [3.10s]
```

## Files changed

- `tests/migrations.test.ts` — improved `startPostgres()` wait strategy + OOM-mitigating docker args
- `tests/migrations.schema.test.ts` — same changes (duplicate helper, both files need the fix)
- `vitest.config.ts` — serialize migration suite to prevent concurrent containers

Closes #57

### DIFF
--- a/tests/migrations.schema.test.ts
+++ b/tests/migrations.schema.test.ts
@@ -58,10 +58,13 @@ async function startPostgres(): Promise<{ container: string; port: number; url: 
       '--rm',
       '--name',
       container,
-      '-e',
-      `POSTGRES_PASSWORD=${PG_PASSWORD}`,
-      '-p',
-      `${port}:5432`,
+      // Faster startup: trust-mode skips md5 auth handshake during initdb.
+      '-e', `POSTGRES_PASSWORD=${PG_PASSWORD}`,
+      '-e', 'POSTGRES_INITDB_ARGS=--auth-host=trust',
+      // Smaller footprint: 128 MB shared_buffers avoids OOM under CI memory pressure.
+      '-e', 'POSTGRES_SHARED_BUFFERS=128MB',
+      '--shm-size=256m',
+      '-p', `${port}:5432`,
       PG_IMAGE,
     ]);
     if (r.code === 0) {
@@ -76,17 +79,25 @@ async function startPostgres(): Promise<{ container: string; port: number; url: 
     throw new Error(`failed to start postgres container: ${lastErr}`);
   }
 
-  const deadline = Date.now() + 30_000;
+  // Wait strategy: first poll docker logs for the canonical "ready to accept connections"
+  // line — this fires only after initdb + WAL recovery completes, eliminating the race
+  // where pg_isready returns 0 during startup but before the postmaster accepts SQL.
+  const deadline = Date.now() + 60_000;
   while (Date.now() < deadline) {
-    const r = dockerRun(['exec', container, 'pg_isready', '-U', 'postgres']);
-    if (r.code === 0) {
-      const url = `postgres://postgres:${PG_PASSWORD}@127.0.0.1:${port}/postgres`;
-      return { container, port, url };
+    const logs = dockerRun(['logs', container]);
+    const combined = logs.stdout + logs.stderr;
+    if (combined.includes('database system is ready to accept connections')) {
+      // Secondary sanity: confirm the port is actually accepting connections.
+      const ready = dockerRun(['exec', container, 'pg_isready', '-U', 'postgres']);
+      if (ready.code === 0) {
+        const url = `postgres://postgres:${PG_PASSWORD}@127.0.0.1:${port}/postgres`;
+        return { container, port, url };
+      }
     }
-    await new Promise((res) => setTimeout(res, 500));
+    await new Promise((res) => setTimeout(res, 300));
   }
   dockerRun(['rm', '-f', container]);
-  throw new Error('postgres container did not become ready in 30s');
+  throw new Error('postgres container did not become ready in 60s');
 }
 
 function stopPostgres(container: string): void {
--- a/tests/migrations.test.ts
+++ b/tests/migrations.test.ts
@@ -47,10 +47,13 @@ async function startPostgres(): Promise<{ container: string; port: number; url: 
       '--rm',
       '--name',
       container,
-      '-e',
-      `POSTGRES_PASSWORD=${PG_PASSWORD}`,
-      '-p',
-      `${port}:5432`,
+      // Faster startup: trust-mode skips md5 auth handshake during initdb.
+      '-e', `POSTGRES_PASSWORD=${PG_PASSWORD}`,
+      '-e', 'POSTGRES_INITDB_ARGS=--auth-host=trust',
+      // Smaller footprint: 128 MB shared_buffers avoids OOM under CI memory pressure.
+      '-e', 'POSTGRES_SHARED_BUFFERS=128MB',
+      '--shm-size=256m',
+      '-p', `${port}:5432`,
       PG_IMAGE,
     ]);
     if (r.code === 0) {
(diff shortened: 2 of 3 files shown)


NikolayS (commented):
## VERDICT: APPROVE (with non-blocking findings)

CI: all three jobs green on commit `a46e9ac` (run 24936701753 — build, aggregator, egress-integration all `success`).

---

### Finding 1 — `POSTGRES_SHARED_BUFFERS` env var is a NO-OP (non-blocking)

**Severity: Low / Cosmetic**

Both test files pass `-e POSTGRES_SHARED_BUFFERS=128MB` as a Docker environment variable. Postgres does **not** read `shared_buffers` from an environment variable — it reads it from `postgresql.conf` or from a `-c` flag passed as the container command. The official `postgres` Docker image has no logic to forward `POSTGRES_SHARED_BUFFERS` into Postgres configuration.

Correct form (if reducing shared_buffers was the goal):
```
'-c', 'shared_buffers=128MB'  // appended after the image name as CMD args
```

The `postgres:16-alpine` default for `shared_buffers` is 128 MB anyway, so the current env var is a no-op that neither helps nor hurts. The OOM concern is more effectively addressed by `--shm-size=256m` (which *is* correctly passed to Docker) and by the Vitest serialization that prevents two containers from running concurrently. Net effect: this is dead config, not a correctness problem.

**Recommendation:** Replace `-e POSTGRES_SHARED_BUFFERS=128MB` with a Docker CMD argument `-c shared_buffers=128MB` in a follow-up, or drop it entirely since 128 MB is already the default.

---

### Finding 2 — Vitest serialization is GLOBAL, not scoped to migration tests (non-blocking, but follow-up warranted)

**Severity: Medium / Performance**

`vitest.config.ts` sets `poolOptions.threads.singleThread: true, maxThreads: 1` at the top level with no file scoping. The `include` glob covers `tests/**`, `apps/*/test/**`, `packages/*/test/**`, and `infra/*/test/**` — meaning every test file in the repo now runs sequentially, including fast unit tests in `packages/llm-adapter`, `apps/web`, etc.

This is the correct fix for the immediate flake, and CI is green, so it is approvable as-is. But as the test suite grows this will become a meaningful CI slowdown.

**Recommendation (follow-up, not a blocker):** Scope serialization to migration tests only. Options in order of preference:
1. Use `vitest.workspace.ts` with a dedicated project for `tests/migrations*.test.ts` that carries `singleThread: true`, leaving all other projects unrestricted.
2. Or use `test.fileParallelism: false` per-file via `// @vitest-environment` pragma + a separate config block.

The PR body acknowledges the OOM concern clearly, CI is passing, and the flake is fixed — this is a pragmatic call. Flagging as a debt item rather than a blocker.

---

### Finding 3 — Local proof used `bun test`, not `bunx vitest run` (non-blocking, CI is source of truth)

**Severity: Informational**

Both test files import from `'vitest'`, not `'bun:test'`. The PR body documents `bun test tests/migrations.test.ts → 3 pass`. With `vitest` imports, `bun test` will not exercise the vitest runner — it likely ran with Bun's native runner falling back silently or skipping vitest-specific APIs. The correct local invocation is `bunx vitest run tests/migrations.test.ts`.

This is non-blocking because CI (which uses `bun run test` → vitest) is the authoritative signal and is green.

---

### What is correct

1. **Wait strategy** — correctly implemented in both files. `startPostgres()` polls `docker logs <container>` for the exact string `"database system is ready to accept connections"`, then issues `pg_isready` as a secondary check. Order is correct (log line first, then port check). Deadline is 60 s. Throws and cleans up on timeout. Poll interval is 300 ms. All good.

2. **`POSTGRES_INITDB_ARGS=--auth-host=trust`** — correctly set as a Docker `-e` env var and confined entirely to test code. Not present in `infra/dev/docker-compose.yml` or any production config. Safe.

3. **`--shm-size=256m`** — correctly passed as a Docker run flag (not a Postgres env). This is the effective OOM mitigation.

4. **`Closes #57`** — present in the PR body and commit message. Issue reference is correct.

5. **No `--no-verify`, no skipped lints** — lint step passes in CI.

6. **No prod config contamination** — `infra/dev/docker-compose.yml` uses `supabase/postgres:15.6.1.146` with no `POSTGRES_INITDB_ARGS`, `POSTGRES_SHARED_BUFFERS`, or `--shm-size` injection.

---

**Summary:** 3 findings — 1 dead config (low), 1 global serialization cost (medium, follow-up), 1 informational. None block merge. The core wait-strategy fix is solid and the flake root cause is correctly addressed.

PR #66, opened 2026-04-25
chore: sweep Sprint 2 review nits (#43 #45 #55 #58 #61 #62)

NikolayS:
## Summary

Six small Sprint 2 review follow-ups bundled into one sweep PR. One commit per issue.

- **#43 — chore(infra): 4 REV non-blocking findings from PR #38**
  - S-NB1: escape single-quotes in migration filename in `is_applied` (SQL injection defence-in-depth; filenames are developer-controlled but categorical safety is cheap).
  - S-NB2: pin `postgres:16-alpine` by SHA256 digest in `scripts/migrate.sh` fallback, `tests/migrations.test.ts`, `tests/migrations.schema.test.ts`, and `apps/api/test/atomic-spend.test.ts` — matches the discipline already applied in `docker-compose.yml`.
  - B-NB1: replace `--network host` with `--add-host=host.docker.internal:host-gateway` in the docker-bundled psql fallback — `--network host` is a no-op shim on Docker Desktop for Mac.
  - B-NB2: add checksum-drift warn in the skip path — when a previously-applied migration's on-disk SHA256 differs from the stored value, emit a `WARN` line to stderr; non-fatal, forensic column stays.

- **#45 — chore(llm-adapter): 3 REV nits from PR #40**
  - Nit 1 (dead-store): remove `lastOutcome` variable — written each loop iteration but never read after the loop exits; replaced with comment.
  - Nit 2 (jitter untested): add inline doc explaining why `Math.random()` is not exercised in CI (tests pass `sleepFn` that resolves immediately) and why it's fine.
  - Nit 3 (close-handler race): add comment near `close` handler explaining the timer-fired-then-clean-exit race and why `settle()` idempotency makes it benign.

- **#55 — chore(api): 3 REV nits from PR #53 (atomic spend)**
  - `atomic-spend.ts` JSDoc: remove stale "clamp" language (clamping was removed in `078c04b`); caller must pass `est_cents ≤ KIND_CEILING[kind]`.
  - `provider-attempts.ts`: drop `est_cents` from `StartAttemptInput` — declared but never written to DB; remove from AC6 test call site too.
  - `atomic-spend.ts`: comment `ReserveSpendResult.ledger_row_id` as intentional null literal — `llm_spend_daily` has no surrogate PK.

- **#58 — feat(api): late_arrivals idempotency — UNIQUE(study_id, visit_id)**
[...]

PR #65, opened 2026-04-25
feat(infra): PGDATA on ZFS + DBLab on willbuy-v01 (#49)

NikolayS:
## Summary

- **`infra/zfs/setup-zfs-pgdata.sh`** — idempotent ZFS pool + dataset creation (`tank/willbuy/pgdata` with `recordsize=8K, compression=lz4, atime=off, logbias=throughput, primarycache=metadata`; optional `tank/willbuy/pgwal` with `recordsize=128K`). Supports `--check` dry-run.
- **`infra/zfs/migrate-pgdata.sh`** — stops Supabase compose stack, rsyncs Docker volume → ZFS mount, writes `docker-compose.zfs.yml` override, restarts and verifies `pg_isready` + schema integrity.
- **`infra/dblab/install-dblab.sh`** — installs DBLab CLI, copies `dblab.yml` config, takes initial ZFS snapshot, starts `postgresai/dblab-server` container, waits for `/healthz`, inits CLI environment.
- **`infra/dblab/dblab.yml`** — DBLab server config template: physical mode + `pg_basebackup`, 6 h snapshot schedule, 4-snapshot retention, clone ports 6000–6019, 240 min idle TTL. Token placeholder replaced at deploy time.
- **`infra/README.md`** — ZFS architecture diagram, exact invocation order, compose override usage, ZFS snapshot backup pattern, DBLab clone usage, CI/per-PR ephemeral database recipe.
- **`SPEC.willbuy.amendments.md`** — A5 follow-on (2026-04-25) documenting pool name, dataset structure, WAL placement decision, migration strategy, and DBLab config summary.

All scripts pass `shellcheck 0.11.0` with zero warnings.

---

## VM disk audit

CPX21 has **80 GB SSD** (Hetzner resource table in `infra/README.md`). The ZFS loopback file is capped at 30 GB (`VDEV_SIZE=30G`), leaving ~50 GB for the OS + Docker layers.

| Layer | Approx. usage |
|-------|---------------|
| Ubuntu 24.04 base + packages | ~5 GB |
| Docker images (Supabase stack) | ~10 GB |
| ZFS loopback pool (PGDATA + WAL) | 30 GB |
| Remaining headroom | ~35 GB |

[...]

Issue #61, opened 2026-04-25
test(infra): scope vitest singleThread to migration tests only (not repo-wide)


NikolayS:
Non-blocking finding from REV review of merged PR #60:

`vitest.config.ts` currently has `poolOptions.threads.singleThread: true, maxThreads: 1` applied **globally** — that means every test in the repo serializes, not just the migration tests that needed it.

This is a 4x+ slowdown for CI runs as the test suite grows. The migration tests are ~3 seconds; a 100-test suite could go from 10s to 40s.

**Fix:** restructure to `vitest.workspace.ts` with a dedicated project for `tests/migrations*.test.ts` that has the singleThread setting; the rest of the workspaces stay parallel.

Lower-priority than #58/#59. File for sweep PR.

Source: rev-pr60 finding #2.

PR #70, opened 2026-04-25
test(infra): extend testcontainer wait-strategy to apps/api tests

NikolayS:
## Summary

**Root cause:** `apps/api/test/atomic-spend.test.ts` had its own inline Postgres startup that was NOT updated by PR #60. It used a `waitReady()` loop that only called `pg_isready` via `docker exec` (which probes the container's Unix socket). This is insufficient for two reasons:
1. `pg_isready` via `docker exec` returns 0 during initdb startup, before the postmaster is ready to execute SQL (the race PR #60 fixed with the log-line wait).
2. `docker exec pg_isready` uses the container's Unix socket, which becomes ready slightly before the host-side Docker NAT port mapping propagates. The subsequent `psql` call from `migrate.sh` connects over the mapped TCP port and hits "server closed the connection unexpectedly".

**Fix:**
- Extracted the inline `startPostgres()` helper from `tests/migrations.test.ts` and `tests/migrations.schema.test.ts` into a shared module at `tests/helpers/start-postgres.ts`.
- The shared helper adds a **three-stage readiness check**:
  1. Poll `docker logs` for `"database system is ready to accept connections"` (PR #60's fix — fires only after initdb + WAL recovery).
  2. Host-side `psql -c 'SELECT 1'` probe to confirm the Docker NAT port mapping is live for external callers. Falls back to `docker exec pg_isready` if `psql` is not installed on the host.
- Updated all four affected test files: `tests/migrations.test.ts`, `tests/migrations.schema.test.ts`, `apps/api/test/atomic-spend.test.ts`, and `apps/api/test/leases.integration.test.ts`.

**Shared helper location:** `tests/helpers/start-postgres.ts`

## Local 5× pass evidence

```
=== Run 1 === Tests  6 passed (6)
=== Run 2 === Tests  6 passed (6)
=== Run 3 === Tests  6 passed (6)
=== Run 4 === Tests  6 passed (6)
=== Run 5 === Tests  6 passed (6)
```

Zero flakes across 5 back-to-back runs (previously ~3/5 would fail).

## Notes for PR #63

PR #63 (`studies.api.test.ts`) has its own inline Postgres startup on a separate branch. Once this PR lands on `main`, PR #63 should rebase and adopt the shared `startPostgres()` helper from `tests/helpers/start-postgres.ts` to pick up the same fix.


NikolayS:
## Merge resolution — main (PRs #65 + #66) merged into branch

### What conflicted
PR #66 (Sprint 2 sweep) added the `S-NB2` digest pin (`postgres:16-alpine@sha256:4e6e670...`) plus dropped `POSTGRES_SHARED_BUFFERS` in `tests/migrations.test.ts`, `tests/migrations.schema.test.ts`, and `apps/api/test/atomic-spend.test.ts`. This PR had in parallel refactored those same files to delegate all Postgres startup to `tests/helpers/start-postgres.ts`.

### Resolution strategy
- **Dropped** main's locally-defined `startPostgres`/`stopPostgres` blocks from `migrations.test.ts`, `migrations.schema.test.ts` (they were being replaced by the helper import).
- **Dropped** main's `PG_IMAGE` const from `atomic-spend.test.ts` (ditto).
- **Applied** the digest pin to `tests/helpers/start-postgres.ts` — all three consumer test files now get `postgres:16-alpine@sha256:4e6e670bb069649261c9c18031f0aded7bb249a5b6664ddec29c013a89310d50` through the single helper.
- `POSTGRES_SHARED_BUFFERS` is absent from the helper (was never added); `--shm-size=256m` remains.
- `vitest.workspace.ts` from #61 incorporated intact.

### Merge commit
`b2478d2` — pushed to `fix/atomic-spend-testcontainer-flake`

### CI run
https://github.com/NikolayS/willbuy/actions/runs/24937796580 (triggered automatically after push, currently running)

### Local 5× pass evidence for `atomic-spend.test` (flake-confirming)
```
Run 1: 6 passed  Duration 1.93s
Run 2: 6 passed  Duration 1.96s
Run 3: 6 passed  Duration 1.90s
Run 4: 6 passed  Duration 1.96s
Run 5: 6 passed  Duration 1.96s
```
All other suites also green:
- `leases.integration.test` — 13 passed
- `tests/migrations.test.ts` — 3 passed
- `tests/migrations.schema.test.ts` — 81 passed

Ready to merge once CI goes green.

PR #14, opened 2026-04-25
feat(api): Fastify skeleton + structured logging + health endpoint

NikolayS:
Closes #6

Implements spec **§4.1** (API server: Node 20+ Fastify TS) and **§5.12** (field-level logging policy: no URLs / share-token / API keys / payloads / a11y-tree / LLM strings / backstory text / emails — URLs hashed with salted SHA-256, API keys masked to last 4).

## What's in

- `apps/api/src/env.ts` — zod env schema. `URL_HASH_SALT` is required and must be ≥ 32 chars; `PORT` defaults to 3000; `LOG_LEVEL` defaults to `info`. `parseEnv()` throws a single error listing every offending field.
- `apps/api/src/logger.ts` — pino instance with a custom `formatters.log` walker. At any nesting depth: `url` is replaced with `url_hash = sha256(salt+url).slice(0,16)`; `api_key` is masked to last 4 (`***1234`); `share_token`, `email`, `backstory`, `a11y_tree`, `llm_output`, `provider_payload`, `password` are dropped. Exports `hashUrl(salt, url)` and `buildLogger`.
- `apps/api/src/server.ts` — `buildServer(env)` returns a Fastify instance with the redacting logger and a GET `/health` route returning `{ status, uptime, version }` (version read from `apps/api/package.json`).
- `apps/api/src/index.ts` — boot: `parseEnv(process.env)` → if it fails, write `[api] boot failed: <message>` to stderr and exit 1; otherwise build + listen, with graceful SIGTERM/SIGINT shutdown.
- `apps/api/Dockerfile` — multi-stage (deps/build/runtime), non-root UID 1001, HEALTHCHECK on `/health`.
- `apps/api/.env.example`.

## TDD evidence

Two-commit history per acceptance criterion:

- `d08340c test(api): red — Fastify skeleton, env zod, /health, redacting logger` — adds the four test files. CI on this commit alone would show 4 failed test files (no `src/*` exists yet).
- `69d44c2 feat(api): green — Fastify skeleton, env zod, /health, redacting logger` — adds the four source files. CI green.

The redactor is the load-bearing part. Tests assert positive behavior (`url_hash` is the right value, `api_key` is `***last4`, forbidden field names are absent) **and** defense-in-depth: each test serializes the full log record and asserts the raw URL host/path/share-token/email/backstory/llm_output/password literal **never appears** anywhere in the JSON.

## Local run output

```
[...]

PR #92, opened 2026-04-25
chore(api): address 3 REV non-blocking observations from PR #14 (#16)

### LINKED ISSUE #16: chore(api): address 3 REV non-blocking observations from PR #14
Three non-blocking findings from REV on PR #14, follow up when convenient:

1. **Dead branch in `maskApiKey`**: `typeof !== 'string'` guard is unreachable given the call site already type-narrows. Either prove it's reachable (and add a test) or delete it.
2. **`WeakSet` mislabels DAGs as `[Circular]`**: The recursive redactor reuses a `WeakSet` across the walk, so a node visited twice via two different paths gets reported as circular even though it isn't. Fix is to use a depth-tracking approach instead of visited-set.
3. **URL replacement misses non-`url` field names**: Currently the redactor only hashes fields literally named `url`. Real-world records carry `redirect_url`, `target_url`, `source_url`, `callback_url`, `webhook_url`, etc. Either expand the matcher to `*_url` or use a value-shape sniffer (regex on `https?://`).

All three were posted as inline comments on PR #14 with file:line references.

NikolayS:
## Summary

- **Fix 1 — Dead branch in `maskApiKey`:** deleted the unreachable `typeof key !== 'string'` guard; the only call site already narrows to `string` before invoking the function. No behaviour change, no new test needed.
- **Fix 2 — WeakSet DAG false-positive:** switched from a global visited-set to an ancestry-only (depth-tracking) `WeakSet` with `finally { ancestry.delete(node) }` backtracking. A DAG node reachable via two paths is no longer falsely reported as `[Circular]`; real self-referential cycles are still caught.
- **Fix 3 — URL hashing extended to all bare-URL values:** instead of matching only the literal field name `url`, the redactor now hashes any string field whose entire value matches `^https?:\/\/\S+$` (option b from the issue). The output key becomes `<fieldName>_hash`. A `description` containing a URL as a substring is intentionally left untouched (full-string match only).

## Test summary

5 new tests in `apps/api/test/redact.test.ts` (TDD red→green):
- DAG shared object is not falsely flagged as `[Circular]`
- Real self-referential cycle (`a.self = a`) is correctly reported as `[Circular]`
- `redirect_url`, `webhook_url`, `target_url` fields are hashed
- `description: "see https://example.com for details"` is NOT hashed
- Arbitrary field (`endpoint`) whose full value is a bare URL is hashed

All 60 tests pass; `tsc --noEmit` clean.

Closes #16

### DIFF
--- a/apps/api/src/logger.ts
+++ b/apps/api/src/logger.ts
@@ -24,39 +24,59 @@ export function hashUrl(salt: string, url: string): string {
     .slice(0, 16);
 }
 
+// Fix 1: removed unreachable `typeof key !== 'string'` branch — the only call
+// site already guards with `typeof v === 'string'` before invoking maskApiKey.
 function maskApiKey(key: string): string {
-  if (typeof key !== 'string') return '***';
   const last4 = key.slice(-4);
   return `***${last4}`;
+}
+
+// Returns true when `value` is a complete URL string (not a substring within
+// a longer sentence). Used by Fix 3 to hash bare-URL values regardless of
+// field name.
+const BARE_URL_RE = /^https?:\/\/\S+$/;
+function isBareUrl(value: unknown): value is string {
+  return typeof value === 'string' && BARE_URL_RE.test(value);
 }
 
 // Recursively rewrite a log payload per the spec §5.12 redaction policy.
 // We do this in a custom serializer (not pino's `redact` paths) because the
 // spec requires field-name based stripping at any nesting depth, including
 // inside arbitrary user-supplied context objects.
-function redact(value: unknown, salt: string, seen: WeakSet<object>): unknown {
+//
+// Fix 2: `ancestry` tracks only the current call stack (depth-first path),
+// not every object ever visited. When we leave a node we remove it from the
+// set (backtracking), so a DAG node reachable via two paths is NOT falsely
+// reported as [Circular] — only a node that is an ancestor of itself is.
+function redact(value: unknown, salt: string, ancestry = new WeakSet<object>()): unknown {
   if (value === null || typeof value !== 'object') return value;
-  if (seen.has(value as object)) return '[Circular]';
-  seen.add(value as object);
-
-  if (Array.isArray(value)) {
-    return value.map((v) => redact(v, salt, seen));
-  }
-
-  const out: Record<string, unknown> = {};
-  for (const [k, v] of Object.entries(value as Record<string, unknown>)) {
-    if (REMOVE_FIELDS.has(k)) continue;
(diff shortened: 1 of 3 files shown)


NikolayS:
## VERDICT: APPROVE

Manager review (rev-pr92 sub-agent quota-blocked; reviewed in main thread, opus-1M).

### What was verified

**Fix 1 — maskApiKey dead branch removed.** `apps/api/src/logger.ts` no longer has `if (typeof key !== 'string')`. Existing call site at line ~84 already type-narrows `typeof v === 'string'` before invoking. Clean.

**Fix 2 — Per-call ancestry WeakSet.**
- Renamed `seen` → `ancestry`. Default param `= new WeakSet<object>()`.
- `try { … recurse … } finally { ancestry.delete(value as object); }` — correct backtracking on every code path (array branch, object branch, throw).
- Recursion passes the same `ancestry` to children. After a child returns, the child is popped, allowing a sibling to traverse the same shared subtree without false `[Circular]`.
- DAG case (shared leaf via two paths) → both copies preserved.
- Real cycle (`a.self = a`) → still `[Circular]` as expected.
- `buildLogger` no longer creates a new WeakSet at the call site (now the default param handles it).

**Fix 3 — Bare-URL value-shape sniffer.**
- `BARE_URL_RE = /^https?:\/\/\S+$/` — anchored both ends, no whitespace, full string match.
- `isBareUrl(v)` typeguard. Replaces `url`-only field-name match.
- Output key is `${k}_hash` — preserves the field name for traceability (e.g., `redirect_url` → `redirect_url_hash`).
- Substring case (`"see https://example.com for details"`) NOT hashed — the regex `^…$` anchors prevent substring matches.

### Test coverage

`apps/api/test/redact.test.ts` (new file, 5 tests) covers:
1. DAG shared object — both copies preserved as `'safe'`.
2. Real cycle — `a.self` becomes `'[Circular]'`.
3. `redirect_url`/`webhook_url`/`target_url` all hashed; original field names disappear; raw hostnames absent from JSON.
4. `description` field with substring URL → unchanged.

(Plus an implicit Fix-1 regression-check via existing logger tests still passing.)

### TDD discipline
- `0fd25e6` RED — tests-only.
- `b6960dc` GREEN — implementation only.
- `8884341` lint fix (URL_FIELD removal cleanup).
- `95637bc` env literal fix (cascade-merge fallout).
- Two cascade merges from main.

Clean separation.

### CI
ALL GREEN (build + aggregator + egress-integration) on `b6fcc11`.

### No findings.

Self-approval blocked → posted as `--comment` per fallback.

PR #95, opened 2026-04-25
feat(web+api): magic-link sign-in (#79)

NikolayS:
## Summary

- `POST /api/auth/magic-link` — validates email, upserts `accounts(owner_email)`, generates 22-char nanoid token, stores `sha256(token)` in `auth_magic_links`, sends via Resend (or returns `verifyUrl` in body when `WILLBUY_DEV_SESSION` is set).
- `GET /api/auth/verify?token=<t>` — timing-safe token lookup, single-use check, issues 7-day HMAC-signed `wb_session` cookie (HttpOnly, SameSite=Lax, `__Host-` prefix in production), 302 → `/dashboard`. Returns 404 for all invalid states (§2 #20 — no existence leak).
- `POST /api/auth/sign-out` — clears cookie (Max-Age=0), 302 → `/sign-in`.
- `apps/web/app/sign-in/page.tsx` — email form with success/error states, CSP-compliant (no inline scripts/styles, §5.10).
- `infra/migrations/0014_auth_magic_links.sql` — `auth_magic_links` table + index on `token_hash` + `UNIQUE` on `accounts.owner_email`.

Session middleware (`apps/api/src/auth/session.ts`) is a sibling to the existing `api-key.ts` — both coexist. API key auth for programmatic clients; session auth for browser users.

## Curl evidence

```
# PATH 1: POST /api/auth/magic-link (dev mode — WILLBUY_DEV_SESSION set)
$ curl -s -X POST http://localhost:3099/api/auth/magic-link \
    -H 'Content-Type: application/json' \
    -d '{"email":"demo@example.com"}'
→ Status: 202
→ Body:   {"verifyUrl":"http://localhost:3099/api/auth/verify?token=WhWo7rj0AjXRJNZMeo-PKk"}

# PATH 3: GET /api/auth/verify?token=<valid>
$ curl -s -I "http://localhost:3099/api/auth/verify?token=WhWo7rj0AjXRJNZMeo-PKk"
→ Status:     302
→ Location:   /dashboard
→ Set-Cookie: wb_session=<...>; HttpOnly; SameSite=Lax; Path=/; Max-Age=604800

# PATH 4: same token reused (already used → 404)
$ curl -s -I "http://localhost:3099/api/auth/verify?token=WhWo7rj0AjXRJNZMeo-PKk"
→ Status: 404

# Invalid email → 400
$ curl -s -X POST http://localhost:3099/api/auth/magic-link \
    -H 'Content-Type: application/json' \
[...]

Issue #26, opened 2026-04-25
feat(api): postgres schema migrations (provider_attempts, leases, ledger, etc.)


NikolayS:
## Sprint 2 — Foundation #2: Postgres schema migrations

**Spec:** §4.1 (canonical table list — `accounts`, `api_keys`, `studies`, `page_captures`, `backstories`, `visits`, `reports`, `credit_ledger`, `llm_spend_daily`), §5.5 (atomic spend reservation row), §5.11 (single-writer finalize uses a `studies.status` flow + `late_arrivals`), §5.13 broker schema → `page_captures` row, §5.15 (`provider_attempts` table with `logical_request_key UNIQUE`), §16 (unified `provider_attempts` ledger).

**Depends on:** S2-1 (migrations runner).

**Outcome:** Every migration named `000N_*.sql` in `infra/migrations/` lands cleanly under `pnpm migrate` and produces the canonical schema. Re-runs are idempotent.

## In scope (split per migration file for reviewability)

- `0001_accounts_and_keys.sql` — `accounts(id, owner_email, created_at)`, `api_keys(id, account_id FK, key_hash, prefix, last_used_at, revoked_at, created_at)`. UNIQUE on hash. Constraint: ≤2 active keys per account.
- `0002_studies.sql` — `studies(id, account_id FK, kind enum {single, paired}, status enum {pending, capturing, visiting, aggregating, ready, failed}, created_at, finalized_at)`.
- `0003_page_captures.sql` — `page_captures(id, study_id FK, side enum {A, B} NULL, url_hash, a11y_storage_key, screenshot_storage_key NULL, host_count, status enum {ok, blocked, error}, breach_reason NULL, captured_at)`. Unique by (study_id, side).
- `0004_backstories_and_leases.sql` — `backstories(id, study_id FK, idx int, payload jsonb, lease_until NULL, lease_owner NULL, heartbeat_at NULL)`. UNIQUE (study_id, idx). `backstory_leases(study_id, backstory_id, lease_until, lease_owner)` per spec §5.11.
- `0005_visits.sql` — `visits(id, backstory_id FK, side enum {A, B}, status enum {started, ok, failed, indeterminate}, parsed jsonb NULL, raw_storage_key NULL, repair_generation int, transport_attempts int, score numeric NULL, started_at, ended_at)`. UNIQUE (backstory_id, side).
[...]

PR #18, opened 2026-04-25
feat(shared): visitor schema + scoring weights + backstory types

NikolayS:
Closes #4

## Summary

- `packages/shared` (`@willbuy/shared`) now exposes `VisitorOutput`, `Backstory`, `NextAction`, `NEXT_ACTION_WEIGHTS`, `PAID_TIERS`, and `scoreVisit` as the single zod-validated source of truth — every other package can import them per the issue's stated outcome.
- `scoreVisit` and `NEXT_ACTION_WEIGHTS` mirror the growth team's `ab/pricing-page-2026apr/scoring.md` rubric verbatim (1.0 / 0.8 / 0.6 / 0.3 / 0.2 / 0.0 with the bookmark-and-paid-tier and start_free_hobby-and-paid-tier bumps); `Backstory` mirrors `personas/backstories.md` dimension by dimension. Every zod field carries a `.describe()` cross-referencing the spec section.
- The growth rubric uses 8 next_action values (paid-conversion-aligned), which extends the 6-value list in spec §2 #15/#18/#19. SPEC amendment **A1 (2026-04-24)** records the deviation per CLAUDE.md "spec is authoritative" rule. McNemar binarization (§2 #19) is updated accordingly: `converted = 1 IFF next_action ∈ {purchase_paid_today, contact_sales, book_demo, start_paid_trial}`.

## Spec link

Implements spec **§2 #15** (visitor output schema) + **§2 #5** (backstory dimensions) + **§5.6 / §2 #18** (conversion-weighted next-action score), as amended by `.samo/spec/willbuy/SPEC.willbuy.amendments.md` **A1 (2026-04-24)**.

## Test evidence — red→green for every acceptance branch

\`pnpm --filter @willbuy/shared test --reporter=verbose\` output (current HEAD):

\`\`\`
 ✓ test/visitor.test.ts > VisitorOutput (spec §2 #15) > parses a known-valid fixture
 ✓ test/visitor.test.ts > VisitorOutput (spec §2 #15) > rejects when a required field is missing
 ✓ test/visitor.test.ts > VisitorOutput (spec §2 #15) > rejects an oversized field (first_impression > 400 chars per §2 #15)
 ✓ test/visitor.test.ts > VisitorOutput (spec §2 #15) > rejects an out-of-range integer (will_to_buy must be 0–10 per §2 #15)
 ✓ test/visitor.test.ts > VisitorOutput (spec §2 #15) > rejects a wrong enum value for next_action
[...]

Issue #12, opened 2026-04-25
chore(monorepo): address 2 REV LOW findings from PR #10


NikolayS:
Both non-blocking nits from REV on #10:

1. `tests/lint-fixtures` should be added to `.prettierignore` so `pnpm format` cannot accidentally re-format a fixture and flip its lint expectation. One-line addition.
2. The `willbuy/no-sandbox-flag` ESLint rule misses runtime-concatenated split literals (e.g. `"--no" + "-sandbox"`). The defense-in-depth answer is that the real boundary is at iptables/seccomp/Chromium-sandbox enabled per spec §5.13, not the lint — but tightening the rule (AST-walking string concatenations) would catch obvious bypasses too. Optional follow-up.

Pick up in a slow moment; not blocking any Sprint 1 work.

PR #96, opened 2026-04-25
feat(capture-worker): production broker wiring + integration test (#84)

### LINKED ISSUE #84: feat(capture-worker): production broker wiring + end-to-end integration test (pending→capturing→visiting)
## Sprint 3 — Wiring #1: capture-worker → broker production wiring

**Spec:** §5.1 (data flow step 5: capture worker leases a job, starts hardened container, sends typed message over Unix socket to broker, broker stores artifacts, flips status to `ok`), §5.13 (capture broker — one-way artifact channel; Unix domain socket at `/run/willbuy/broker.sock`), §5.11 (study status transitions: `pending` → `capturing` → `visiting`).

**Depends on:** #42 (capture-worker merged; `run-with-netns.ts` exists), #52 (Bun cutover), #48 (sqlever migrations, for a clean schema baseline).

**Outcome:** Running `bun run start` in `apps/capture-worker` in a production-like environment actually: (a) polls the job queue for pending capture jobs, (b) spawns a hardened container via `run-with-netns.ts` with the broker socket bind-mounted, (c) the container runs Playwright, writes artifacts to the broker socket, (d) the broker flips the visit status to `ok` / `blocked`, (e) the study transitions from `capturing` to `visiting` when all capture jobs for the study are done. An integration test verifies the full path against a fixture page without requiring a live LLM.

## In scope

- `apps/capture-worker/src/index.ts` — wire the polling loop to call `run-with-netns.ts`; currently the entrypoint exists but has no production job-queue polling.
- `apps/capture-worker/src/broker-client.ts` (or equivalent) — the in-container side that writes the typed artifact message to `/run/willbuy/broker.sock` after capture completes; validated against the broker schema from `apps/api/src/broker/schema.ts` (or wherever it lives).
- `apps/api/src/broker/server.ts` — ensure the broker socket server is started as part of the API process (or as a standalone `apps/broker` service if that's cleaner); currently unclear if it's wired in production.
- Job-queue poll: `SELECT … FOR UPDATE SKIP LOCKED` with 90 s lease (per spec §2 #29); heartbeat every 20 s; DLQ after `max_attempts=5`.
- Study status transition: when ALL visit rows for a study have `status IN ('ok', 'blocked')`, transition study to `visiting` (kicks visitor-worker polling).
- Integration test `apps/capture-worker/test/broker-integration.test.ts`:
  - Starts a local fixture HTTP server (static HTML page).
  - Inserts a study + visit row into a test DB.
  - Runs capture worker against that job.
  - Asserts: visit row → `ok`; `page_captures` row exists with non-empty `a11y_tree_b64`; study status → `visiting`.
  - Does NOT require a real LLM.

## TDD acceptance

1. Capture worker polls queue, leases a job, runs container, broker receives artifact → visit status `ok`.
2. Blocked page (fixture returns `__cf_bm` cookie pattern) → visit status `blocked`; study eventually transitions to `failed` if all visits blocked.
3. Container exits with non-zero code (simulate crash) → lease released; job re-queued; `attempt_count` incremented.
4. Job exceeds `max_attempts=5` → moved to DLQ; study `failed`.
5. Study with all visits `ok` → study status transitions to `visiting`.
6. Broker socket receives oversized artifact → rejected; visit marked `blocked`; no data persisted.

## Out of scope

- Visitor-worker integration (covered by #59).
- Firecracker microVM (v0.2).
- Production deploy / Docker Compose wiring (captured in deployment runbook; not this issue).
- Full end-to-end with LLM (dogfood issue #87).

## Definition of done

Standard PR contract. Integration test green in CI. Manual test on willbuy-v01 staging: create study via `POST /studies`, watch visit status transition through `pending` → capture job picked up → `ok` → study → `visiting` (confirm via `GET /studies/:id` poll or DB query).

NikolayS:
## Summary

Implements issue #84: production wiring of capture-worker → broker, plus an end-to-end integration test that verifies `pending → capturing → visiting` transitions through the full Postgres + broker + capture-container path.

Per the issue body and spec §5.1 step 5 / §5.13 / §5.11.

## Test plan
- [x] Integration test passes locally
- [x] CI green (build/aggregator/egress-integration)

Closes #84

### DIFF
--- a/apps/api/src/routes/studies.ts
+++ b/apps/api/src/routes/studies.ts
@@ -143,12 +143,16 @@ export async function registerStudiesRoutes(
         }
 
         // Create study row.
+        // urls[]: persisted so capture-worker can read studies.urls[variant_idx]
+        // when leasing a visit (issue #84, PR #96 B3 fix). The 1..2 cardinality
+        // CHECK constraint mirrors the zod schema above; the column was added
+        // in migration 0016_studies_urls.sql.
         const kind = body.urls.length === 2 ? 'paired' : 'single';
         const studyResult = await client.query<{ id: string }>(
-          `INSERT INTO studies (account_id, kind, status)
-           VALUES ($1, $2, 'capturing')
+          `INSERT INTO studies (account_id, kind, status, urls)
+           VALUES ($1, $2, 'capturing', $3::text[])
            RETURNING id`,
-          [String(account.id), kind],
+          [String(account.id), kind, body.urls],
         );
         const studyId = studyResult.rows[0]!.id;
 
--- a/apps/capture-worker/package.json
+++ b/apps/capture-worker/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "build": "tsc -p tsconfig.json --noEmit",
     "typecheck": "tsc -p tsconfig.json --noEmit --pretty",
-    "test": "vitest run --exclude test/smoke.test.ts --exclude test/integration.test.ts",
+    "test": "vitest run --exclude test/smoke.test.ts",
     "test:integration": "vitest run test/integration.test.ts",
     "smoke": "vitest run test/smoke.test.ts"
   },
--- a/apps/capture-worker/package.json
+++ b/apps/capture-worker/package.json
@@ -8,15 +8,20 @@
   "scripts": {
     "build": "tsc -p tsconfig.json --noEmit",
     "typecheck": "tsc -p tsconfig.json --noEmit --pretty",
-    "test": "vitest run --exclude test/smoke.test.ts",
+    "test": "vitest run --exclude test/smoke.test.ts --exclude test/integration.test.ts",
+    "test:integration": "vitest run test/integration.test.ts",
     "smoke": "vitest run test/smoke.test.ts"
   },
   "dependencies": {
+    "pg": "^8.20.0",
     "playwright": "1.45.0"
   },
   "devDependencies": {
     "@types/node": "^22.10.2",
+    "@types/pg": "^8.20.0",
+    "@willbuy/capture-broker": "workspace:*",
     "typescript": "^5.7.2",
-    "vitest": "^2.1.8"
+    "vitest": "^2.1.8",
+    "zod": "^3.23.0"
   }
 }
--- a/apps/capture-worker/src/broker-client.ts
+++ b/apps/capture-worker/src/broker-client.ts
@@ -1,0 +1,157 @@
+/**
+ * broker-client.ts — Unix-socket client for the capture broker (spec §5.13).
+ *
+ * Writes one typed {@link CaptureRequest} message over the broker Unix domain
+ * socket and reads back the {@link BrokerAck} response.
+ *
+ * Protocol: length-prefixed framing identical to the broker server side
+ * (apps/capture-broker/src/framing.ts). Single shot per connection.
+ *
+ * This module intentionally duplicates the framing constants rather than
+ * importing from capture-broker (separate package; we don't want a hard dep
+ * on the broker's internals from the worker).
+ */
+
+import { connect } from 'node:net';
+
+// ── framing ───────────────────────────────────────────────────────────────────
+// Mirrors apps/capture-broker/src/framing.ts — 4-byte big-endian length prefix.
+
+const HEADER_BYTES = 4;
+
+function frame(payload: Buffer): Buffer {
+  const header = Buffer.alloc(HEADER_BYTES);
+  header.writeUInt32BE(payload.length, 0);
+  return Buffer.concat([header, payload]);
+}
+
+// ── types (re-stated from apps/capture-broker/src/schema.ts) ─────────────────
+// We re-declare the wire types here so capture-worker has no package-level
+// dep on capture-broker. The broker's schema.ts is the source of truth;
+// any drift here will be caught by the integration test.
+
+export type CaptureRequestPayload = {
+  status: 'ok' | 'blocked' | 'error';
+  a11y_tree_b64: string;
+  screenshot_b64?: string;
+  banner_selectors_matched: string[];
+  overlays_unknown_present: boolean;
+  blocked_reason?: string;
+  host_count: number;
+  breach_reason?: string;
+};
+
+export type BrokerAck =
+  | {
+      ok: true;
+      capture_id: string;
+      a11y_object_key: string;
+      screenshot_object_key?: string;
+    }
+  | {
+      ok: false;
+      error: string;
+      detail?: string;
+    };
+
+export type BrokerClientOpts = {
+  /** Path to the Unix domain socket (default: /run/willbuy/broker.sock). */
+  socketPath?: string;
+  /** Connection + read timeout in ms (default: 30_000). */
+  timeoutMs?: number;
+};
+
+const DEFAULT_SOCKET_PATH = '/run/willbuy/broker.sock';
+const DEFAULT_TIMEOUT_MS = 30_000;
+
+/**
+ * Send a single capture artifact message to the broker and await the ack.
+ *
+ * Throws on connection failure, timeout, or framing error. Returns the
+ * structured {@link BrokerAck} (which may itself be `ok: false` for
+ * schema / byte-cap rejections).
+ */
+export async function sendToBroker(
+  payload: CaptureRequestPayload,
+  opts?: BrokerClientOpts,
+): Promise<BrokerAck> {
+  const socketPath = opts?.socketPath ?? DEFAULT_SOCKET_PATH;
+  const timeoutMs = opts?.timeoutMs ?? DEFAULT_TIMEOUT_MS;
+
+  return new Promise<BrokerAck>((resolve, reject) => {
+    const socket = connect(socketPath);
+
+    let settled = false;
+    const chunks: Buffer[] = [];
+    let totalBytes = 0;
+
+    const timer = setTimeout(() => {
+      if (!settled) {
+        settled = true;
+        socket.destroy();
+        reject(new Error(`broker-client: timeout after ${timeoutMs}ms`));
+      }
+    }, timeoutMs);
+
+    const finish = (result: BrokerAck): void => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      resolve(result);
+    };
+
+    const fail = (err: Error): void => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      reject(err);
+    };
+
+    socket.on('error', fail);
+
+    socket.on('connect', () => {
+      const body = Buffer.from(JSON.stringify(payload), 'utf8');
+      const framed = frame(body);
+      // Write message then half-close (FIN) — broker expects this.
+      socket.write(framed, (writeErr) => {
+        if (writeErr) {
+          fail(writeErr);
+          return;
+        }
+        socket.end();
+      });
+    });
+
+    socket.on('data', (chunk: Buffer) => {
+      chunks.push(chunk);
+      totalBytes += chunk.length;
+    });
+
+    socket.on('end', () => {
+      // Reassemble and parse the framed response.
+      const buf = Buffer.concat(chunks, totalBytes);
+      if (buf.length < HEADER_BYTES) {
+        fail(new Error(`broker-client: short response (${buf.length} bytes)`));
+        return;
+      }
+      const declaredLen = buf.readUInt32BE(0);
+      if (buf.length < HEADER_BYTES + declaredLen) {
+        fail(
+          new Error(
+            `broker-client: truncated response (got ${buf.length} bytes, expected ${HEADER_BYTES + declaredLen})`,
+          ),
+        );
+        return;
+      }
+      const payloadBuf = buf.slice(HEADER_BYTES, HEADER_BYTES + declaredLen);
+      let ack: BrokerAck;
+      try {
+        ack = JSON.parse(payloadBuf.toString('utf8')) as BrokerAck;
+      } catch (e) {
+        fail(new Error(`broker-client: unparseable ack: ${e instanceof Error ? e.message : String(e)}`));
+        return;
+      }
+      finish(ack);
+    });
+  });
+}
--- a/apps/capture-worker/src/broker-client.ts
+++ b/apps/capture-worker/src/broker-client.ts
@@ -112,7 +112,20 @@ export async function sendToBroker(
     socket.on('connect', () => {
       const body = Buffer.from(JSON.stringify(payload), 'utf8');
       const framed = frame(body);
-      // Write message then half-close (FIN) — broker expects this.
+      // Lifecycle (PR #96 M1 fix):
+      //   1. Write the framed request, then half-close (FIN) so the broker's
+      //      `readOneFrame` loop wakes up. The broker waits for a trailing
+      //      sentinel byte OR FIN to confirm single-shot framing — without
+      //      a half-close it would block until its frame timeout.
+      //   2. Continue listening on the readable side. socket.end() only
+      //      shuts down the WRITABLE half; the readable half keeps receiving
+      //      ack bytes from the broker.
+      //   3. Resolve as soon as we have parsed the full length-prefixed ack
+      //      (do NOT wait for the broker's 'end'). This removes the
+      //      dependency on FIN ordering that the reviewer flagged in M1.
+      //   4. Destroy the socket once parsed. 'end' / 'close' are still hooked
+      //      as failure paths if the broker drops the connection without
+      //      writing a complete ack.
       socket.write(framed, (writeErr) => {
         if (writeErr) {
           fail(writeErr);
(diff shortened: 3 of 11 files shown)


NikolayS:
## VERDICT: REQUEST CHANGES

**Reviewer:** rev-pr96 (read-only audit; did not run local tests)
**CI:** all 3 checks green (build / aggregator / egress-integration). **But see Finding 1 — CI does not actually exercise the broker wiring.**
*(self-approval blocked → posting via `--comment` fallback)*

### Findings (3 blockers, 3 medium/low, 2 notes)

**1. [BLOCKER] CI never runs `test/integration.test.ts`.**
`apps/capture-worker/package.json` `test` script: `vitest run --exclude test/smoke.test.ts --exclude test/integration.test.ts`. No CI job invokes `test:integration`. The "ALL GREEN" status reflects only the unit subset; the broker-client ack-parse path is **unverified by CI**. This is exactly the scenario where a Bun socket-lifecycle regression (or any other broker-wiring bug) ships unnoticed. Either (a) re-include `integration.test.ts` in the standard `test` script (gated on availability of Docker for the Postgres testcontainer), or (b) add a dedicated CI job that runs `bun run --filter @willbuy/capture-worker test:integration`.

**2. [BLOCKER] Non-durable lease — duplicate captures under concurrency.**
`pollOnce()` does `SELECT … FOR UPDATE SKIP LOCKED` then **immediately `COMMIT`s** before running the capture (lines 105–106 of `poller.ts`). The visit row stays at `status='started'`, so a second concurrent worker will re-lease and re-process the same visit, producing duplicate broker artifacts and possible duplicate `page_captures` rows. The comment "the FOR UPDATE lock is the mutex" is incorrect — the lock is released as soon as the transaction commits. Fix: transition `status='started' → 'leased'` (or equivalent) inside the same transaction before committing, OR keep the long transaction open for the capture duration (the existing 45s ceiling is well under the idle-in-transaction timeout per the poller's own header comment).

**3. [BLOCKER] Production entrypoint silently captures `about:blank`.**
`index.ts` calls `runPollingLoop({ pool, signal, brokerSocketPath })` with no `targetUrlOverride`, and the schema has no `studies.urls`. Result: `targetUrl=''` for every visit → falls into the `opts.skipCapture || !targetUrl` synthetic-artifact branch → broker receives a single-node `WebArea` a11y tree for `about:blank` for every visit in production. The poller comment acknowledges this ("production workers receive it via env/config injection. See follow-up issue for the `studies.urls jsonb` migration"), but the entrypoint ships without that follow-up landed and without a runtime guard. Either gate `index.ts` on a real URL source or fail-fast when none is configured.

**4. [MEDIUM] Bun `socket.write(framed); socket.end();` is unverified end-to-end.**
Per the engineer note in PR context: naive `write+end` can cause Bun's `node:net` to fire local `'end'` before server data arrives. In practice this works here because (a) the broker uses `createServer({ allowHalfOpen: true })` and (b) the broker writes its ack via `socket.end(frame(ack))` so client receives `data+end` together. But Finding 1 means CI is not catching any future regression. Recommend: instead of relying on `'end'` to know the ack has fully arrived, parse based on the length prefix (already declared in the ack header) and `socket.destroy()` once `received >= HEADER_BYTES + declaredLen`. That removes the dependency on FIN ordering. The framing helper used by `readOneFrame` server-side does exactly this; mirror that pattern client-side.

**5. [LOW] Lockfile drift.**
`bun.lock` lists `@types/pg`, `@willbuy/capture-broker`, and `zod` under `dependencies` for `@willbuy/capture-worker`, while `package.json` has them under `devDependencies`. Probably benign but worth a re-`bun install` to align.

**6. [LOW] PR body / macOS guidance.**
PR body doesn't mention `RUN_WITH_NETNS=skip` or the `PLAYWRIGHT_CAPTURE=1` opt-in for macOS dev. The integration test gates on `PLAYWRIGHT_CAPTURE`, not on `RUN_WITH_NETNS`. Worth a one-liner in the PR description so future contributors know how to run the test locally on macOS.

**7. [NOTE — surfacing, not blocking] Visit status semantics drift.**
Spec §5.3 says `Visit: pending → leased → ok | failed`. Schema (`0005_visits.sql`) constrains to `('started','ok','failed','indeterminate')`. PR uses `'started'` as both "queued" and "leased" — pre-existing schema reality, acknowledged in the poller comment. Worth a follow-up issue to reconcile.

**8. [NOTE — clean]**
- TDD discipline: RED `6626d18` (tests + dev deps only) → GREEN `a15670e` (impl). Correct order.
- No `--no-verify`, no `--no-gpg-sign`, no `[skip ci]`.
- No new §2 #12 forbidden identifiers (`conversation_id`, `session_id`, `thread_id`, `previous_response_id`, `cached_prompt_id`, `parent_message_id`, `context_id`, `assistant_id`, `run_id`) introduced.
- No changes to `apps/capture-broker/**` — no regression risk there.

### Summary
The implementation shape is correct (FOR UPDATE SKIP LOCKED, SIGTERM drain, 5s empty-poll backoff, length-prefixed framing mirroring the broker, study `capturing → visiting` transition once all visits terminal). But the three blockers above mean this PR cannot ship as-is: the production entrypoint would publish meaningless captures, the lease is racy, and CI provides no real signal on the broker wiring.

NikolayS:
All 4 findings from the rev-pr96 review are addressed in 300fac1.

**B1 — CI now exercises the integration test.**
`apps/capture-worker/package.json` no longer excludes `test/integration.test.ts` from the standard `test` script. The CI build job ran the suite end-to-end (Postgres + broker + capture-worker) on this push: see `apps/capture-worker/test/integration.test.ts (3 tests) 4289ms` in the build job log. The `test:integration` alias is preserved for manual / focused runs.

**B2 — Lease is durable.**
`pollOnce()` keeps the `SELECT … FOR UPDATE SKIP LOCKED` transaction OPEN across the capture+broker write, so the row lock is held for the entire work unit. `idle_in_transaction_session_timeout=90s` (2× the 45s wall-clock ceiling) bounds a wedged worker. The study-advance transition runs in a separate, post-commit transaction so we never hold two row locks simultaneously. New regression test (`B2: concurrent pollOnce calls process distinct visits`) fires two concurrent pollers and asserts no `visitId` is processed twice.

**B3 — Production reads target URL from the DB.**
Added migration `0016_studies_urls.sql` (matching `infra/sqlever/deploy/0016_studies_urls.sql`) that adds `studies.urls text[]` per spec §4.3. `POST /studies` now persists the URLs the API already received in its zod-validated body. `pollOnce()` reads `s.urls[v.variant_idx + 1]` (1-based PG indexing) joined with the visit row in the lease query. Production is no longer silently capturing `about:blank` for every visit; if neither override nor DB url is configured the visit fails fast (no broker write) with a structured log line. New regression test (`B3: reads target URL from studies.urls[variant_idx]`) exercises the no-override path through the DB.

**M1 — Broker socket lifecycle hardened.**
`broker-client.ts` is now length-prefix-driven: as soon as `HEADER_BYTES + declaredLen` bytes have arrived, we resolve and `socket.destroy()`, no longer relying on the broker's FIN ordering for ack delivery. We still `socket.write()` then half-close (FIN) on the write side because the broker's `readOneFrame` waits for FIN-or-trailing-byte to confirm single-shot framing — that contract is broker-side, not ours to break. The reviewer's concern about Bun's `'end'` discarding in-flight data was addressed by parsing on `'data'` events using the length prefix; `'end'` and `'close'` are now only failure paths.

**Cleanup.** Dropped the 5 untracked scratch files (`broker-debug{,2,3}.ts`, `demo-db-check.ts`, `minimal-socket-test.ts`) that were eng-84's mid-debug artifacts.

**Local verification.**
- `bun --cwd apps/capture-worker typecheck` — clean.
- `bun --cwd apps/capture-worker test` — 20 tests, all passing (was 17; +3 from the integration suite now running by default).
- `bun --cwd apps/api test` — 65 tests, all passing.
- `bunx vitest run tests/migrations.test.ts tests/migrations.schema.test.ts` — 84 passing.

**CI.** Green: https://github.com/NikolayS/willbuy/actions/runs/24941484448 (build / aggregator / egress-integration). The `build` job's `Test` step now includes `apps/capture-worker/test/integration.test.ts (3 tests) 4289ms` — the broker wiring is finally exercised by CI.

Ready for re-review.

NikolayS (commented):
## VERDICT: APPROVE

**Reviewer:** rev-pr96-v2 (read-only re-audit; did not run local tests)
**CI:** all 3 checks green (build / aggregator / egress-integration) on run 24941484448. The `build` job's `Test` step log confirms `apps/capture-worker/test/integration.test.ts (3 tests) 4289ms` actually ran.
*(self-approval blocked → posting via `--comment` fallback; manager merges)*

### Re-verification of the 4 prior findings

**B1 — CI now exercises the integration test. CONFIRMED FIXED.**
`apps/capture-worker/package.json:11` is now `"test": "vitest run --exclude test/smoke.test.ts"` — `--exclude test/integration.test.ts` is gone. CI build-job log on run 24941484448 contains `apps/capture-worker/test/integration.test.ts (3 tests) 4289ms`, confirming the broker wiring is exercised end-to-end on every push. `test:integration` alias preserved for focused local runs.

**B2 — Durable lease. CONFIRMED FIXED.**
`apps/capture-worker/src/poller.ts` `pollOnce()`:
- `BEGIN` (line 97) → `SET LOCAL idle_in_transaction_session_timeout = '90s'` (line 102) → `SELECT v.id, v.study_id, v.variant_idx, s.urls[v.variant_idx + 1] AS study_url FROM visits v JOIN studies s ON s.id = v.study_id WHERE v.status='started' AND s.status='capturing' ORDER BY v.id LIMIT 1 FOR UPDATE OF v SKIP LOCKED` (lines 110–121) → capture + broker write inside the transaction → `UPDATE visits SET status=$1, ended_at=now() WHERE id=$2` (lines 197–200) → `COMMIT` (line 201). The row lock is held for the entire work unit; the 90 s `idle_in_transaction_session_timeout` is exactly 2× the 45 s `CAPTURE_CEILINGS.WALL_CLOCK_MS` per the doc.
- Study-advance transition (`UPDATE studies SET status='visiting' WHERE id=$1 AND status='capturing'`) is moved into `maybeAdvanceStudy()` — a fresh `BEGIN`/`COMMIT` (lines 240–268) on a separate `pool.connect()` (lines 216–221), called only after the lease transaction has committed. No two row locks held simultaneously.
- Regression test `B2: concurrent pollOnce calls process distinct visits` (`integration.test.ts:283–344`) seeds two visits, fires two `pollOnce` calls in `Promise.all`, and asserts `new Set(ids).size === ids.length` — i.e. no `visitId` processed twice.

**B3 — DB-driven URL. CONFIRMED FIXED.**
- Migration `infra/migrations/0016_studies_urls.sql`: `alter table studies add column if not exists urls text[]` + `studies_urls_cardinality_chk check (urls is null or (array_length(urls, 1) between 1 and 2))` per spec §2 #18. `IF NOT EXISTS` + the conditional constraint guard make this idempotent. The nullable + permissive CHECK on legacy rows is the right pattern (legacy rows pre-#84 have NULL).
- Deploy script `infra/sqlever/deploy/0016_studies_urls.sql`: wrapped in `BEGIN; … COMMIT;` with the standard `INSERT INTO _migrations (filename, checksum, applied_at) VALUES ('0016_studies_urls.sql', 'sqlever-managed', NOW()) ON CONFLICT (filename) DO NOTHING` — matching the post-#97-hotfix sqlever pattern.
- `infra/sqlever/sqitch.plan` line 21 adds `0016_studies_urls [0015_seed_sqlever_state] 2026-04-25T22:00:00Z … # persist target URLs on studies (issue #84, PR #96 B3 fix)` — correctly chained after 0015.
- `apps/api/src/routes/studies.ts` line 152–156: `INSERT INTO studies (account_id, kind, status, urls) VALUES ($1, $2, 'capturing', $3::text[])` — persists the zod-validated `urls` array.
- Zod still validates `urls: z.array(z.string().url()).min(1).max(2)` (line 57).
- `poller.ts` lease query (line 114) joins visits to studies and reads `s.urls[v.variant_idx + 1]` — Postgres array indexing is 1-based, `variant_idx` is 0-based externally, so `variant_idx=0 → urls[1]` and `variant_idx=1 → urls[2]`. Correct.
- No-URL fail-fast (lines 150–158): structured `console.error` line + `noUrlFailure=true` → `visitStatus='failed'` (line 165) → no broker write → terminal `UPDATE visits SET status='failed'` inside the lease transaction. No more silent `about:blank`.
- Regression test `B3: reads target URL from studies.urls[variant_idx]` (`integration.test.ts:242–277`) updates `studies.urls` and runs `pollOnce` with no `targetUrlOverride`, asserting the visit reaches `'ok'`.

**Engineer's `studies.urls`-existence claim — VERIFIED CORRECT.**
Read both `infra/migrations/0001_accounts_and_keys.sql` (no `studies` references) and `infra/migrations/0002_studies.sql`. The latter creates `studies (id, account_id, kind, status, created_at, finalized_at)` and explicitly comments at lines 12–16: "*spec §4.3 lists additional columns for studies: urls[], authorization_mode, icp_id, n, seed, screenshots_enabled, cost_cents. These are intentionally omitted from this migration — issue #26 scoped the schema batch to infrastructure tables only.*" The engineer's call to add migration 0016 was unavoidable; the prior task instructions were wrong.

**M1 — Bun socket lifecycle. CONFIRMED FIXED.**
`apps/capture-worker/src/broker-client.ts`:
- `'connect'` handler (lines 112–136): `socket.write(framed, callback)`; on flush, `socket.end()` to half-close the write side. The half-close is required by the broker's `readOneFrame` single-shot framing contract — broker-side concern, correctly preserved.
- `tryParse()` (lines 138–155): reads the 4-byte big-endian length prefix as soon as `totalBytes >= HEADER_BYTES`, then parses the JSON ack body once `totalBytes >= HEADER_BYTES + declaredLen`, then `socket.destroy()` and `finish(ack)`. No longer waits for FIN.
- `'data'` handler (lines 157–164): accumulates chunks and calls `tryParse()` after each chunk. This is the M1 hardening.
- `'error'` handler (line 110): rejects via `fail`.
- `'end'` (lines 166–176) and `'close'` (lines 178–185) handlers: try one final parse, otherwise reject with a descriptive error if data was incomplete. Correct failure paths.

### Other checks

- **TDD discipline:** `git log feat/84-capture-worker-wiring ^origin/main` → `6626d18` RED (tests) → `a15670e` GREEN (impl) → `c8a0a49` merge from main → `300fac1` fix commit (single commit covering all 4 findings). Single fix commit is acceptable here; nit-level preference would be 4 commits but this is fine.
- **No `--no-verify` / no skipped lint:** confirmed in commit metadata; no `[skip ci]` markers.
- **No new §2 #12 forbidden identifiers** (`conversation_id`, `session_id`, `thread_id`, `previous_response_id`, `cached_prompt_id`, `parent_message_id`, `context_id`, `assistant_id`, `run_id`) introduced in any changed file.
- **No regressions to capture-broker package:** `git diff origin/main...feat/84-capture-worker-wiring -- apps/capture-broker/ packages/capture-broker/` is empty. Broker untouched.
- **Scratch-file cleanup:** `git ls-tree -r feat/84-capture-worker-wiring -- apps/capture-worker/test/` shows the 5 scratch files (`broker-debug.ts`, `broker-debug2.ts`, `broker-debug3.ts`, `demo-db-check.ts`, `minimal-socket-test.ts`) are no longer in the tree. Confirmed dropped.

### Nits (non-blocking)

- **PR body still doesn't carry the manual evidence** (psql before/after, `RUN_WITH_NETNS=skip` macOS note, integration-test stdout). The engineer's reply comment has all of this, which is fine for the re-review record, but a future contributor reading just the PR description would still miss the macOS dev guidance. Not a blocker.
- **One commit covering 4 findings** rather than 4 separate commits. Acceptable; scope is narrow and the message clearly enumerates each finding.

### Summary

All 3 blockers and the 1 medium from the prior review are fully addressed. CI green, integration test demonstrably running in CI, lease durability proven by a concurrent regression test, DB-driven URL with fail-fast safety, broker socket lifecycle no longer dependent on FIN ordering. Engineer's `studies.urls` migration call was correct — the prior task instructions were mistaken about the column already existing. Ready to merge.

NikolayS:
## Merged main + resolved 0016 collision (#95 ↔ #96)

Same migration-number-collision pattern as the #97 hotfix — third occurrence. PR #95 (magic-link auth) merged to main first (mergedAt 2026-04-25 21:44 UTC) carrying `0016_auth_magic_links.sql`, so `studies_urls` was renumbered to `0017`.

### Changes in merge commit `5cb9957`
- `infra/migrations/0016_studies_urls.sql` → `0017_studies_urls.sql`
- `infra/sqlever/deploy/0016_studies_urls.sql` → `0017_studies_urls.sql` (incl. file-header comment + `_migrations` INSERT literal)
- `infra/sqlever/sqitch.plan`: `0017_studies_urls` now depends on `0016_auth_magic_links` (chronological order preserved)
- Comment-only updates in `apps/api/src/routes/studies.ts` and `apps/capture-worker/test/integration.test.ts` to reference the new filename
- `seedStudyAndVisit()` in the capture-worker integration test now generates a unique `owner_email` per call: PR #95 added a UNIQUE index on `accounts.owner_email`, which broke the previous shared `'test@willbuy.test'` literal — only surfaced post-merge.

### CI: all green
https://github.com/NikolayS/willbuy/actions/runs/24941729686

| job | result |
|---|---|
| aggregator | pass (37s) |
| build | pass (2m24s) |
| egress-integration | pass (25s) |

### Local verification (single-file scope, no root tests)
- `bunx vitest run tests/migrations.test.ts` — 3/3 passing (count assertion = 17 migrations)
- `bun --cwd apps/capture-worker test integration` — 3/3 passing

Ready to merge.

PR #91, opened 2026-04-25
chore(infra): migrate schema runner from bash to NikolayS/sqlever (#48)

NikolayS:
## Summary

- Replace the ad-hoc bash `scripts/migrate.sh` runner with [NikolayS/sqlever](https://github.com/NikolayS/sqlever) v0.3.0 (amendment A4, issue #48)
- Add sqlever project (`infra/sqlever/`) with `sqitch.conf`, `sqitch.plan`, and 15 deploy scripts (0000–0014)
- `scripts/migrate.sh` becomes a thin wrapper calling `bunx sqlever deploy --top-dir infra/sqlever/ --db-uri $DATABASE_URL --no-tui`
- Add `infra/migrations/0014_seed_sqlever_state.sql` tombstone to keep file counts in sync
- Add `sqlever: "0.3.0"` to `devDependencies`; installed via `bun install --frozen-lockfile` in CI

## sqlever version pinned

**`sqlever 0.3.0`** — installed as `devDependencies["sqlever": "0.3.0"]` in `package.json`, resolved via `bunx sqlever` in `scripts/migrate.sh`.

## State-migration approach

sqlever uses the Sqitch-compatible `sqitch.*` tracking schema (`sqitch.changes`, `sqitch.events`, `sqitch.projects`). The old `_migrations(filename, checksum, applied_at)` table is **kept as a backward-compat shadow** — each deploy script in `infra/sqlever/deploy/` appends `INSERT INTO _migrations ... ON CONFLICT DO NOTHING` inside its `BEGIN`/`COMMIT` block.

For databases previously managed by the old bash runner: `sqlever deploy` is idempotent — all 15 deploy scripts use `IF NOT EXISTS`/`ON CONFLICT DO NOTHING`, so re-running against an existing schema is a no-op. sqlever self-seeds its `sqitch.*` tracking on the first deploy.

State-migration file: [`infra/migrations/0014_seed_sqlever_state.sql`](infra/migrations/0014_seed_sqlever_state.sql) — tombstone comment file added to keep the file count in `infra/migrations/` equal to `_migrations` row count (test assertion).

## CI workflow diff summary

- `bun install --frozen-lockfile` now also installs sqlever (new devDependency; no extra CI step needed)
- Added comment in the `Pre-pull Postgres image` step documenting that `scripts/migrate.sh` delegates to sqlever 0.3.0
- All migration invocations continue through `bun run test` → `scripts/migrate.sh` → `bunx sqlever deploy`

## sqlever project layout

```
[...]

PR #89, opened 2026-04-25
feat(api): §5.12 share-token HttpOnly cookie redirect (#76)

NikolayS:
## Summary

- `GET /reports/:slug?t=<token>` now validates against `share_tokens` table (timing-safe SHA-256 compare), then sets `willbuy_share_<slug>=<HMAC-opaque>; Path=/r/<slug>; HttpOnly; Secure; SameSite=Lax` and 302-redirects to the bare URL.
- `GET /reports/:slug` with the cookie re-validates the HMAC **and** re-checks `revoked_at`/`expires_at` in DB on every request (revoke propagation within one request).
- All invalid/expired/revoked/tampered paths return 404 (no 401, no existence leak per §2 #20).
- `SHARE_TOKEN_HMAC_KEY` added to `apps/api/src/env.ts` (dev default present; production value from 1Password).

## Curl evidence (paths 1, 2, 5, 8 — via `app.inject()` in integration tests)

**Path 1 — valid token in URL → 302 + Set-Cookie: HttpOnly:**
```
# Simulated via app.inject (equivalent to):
curl -v "http://localhost:3000/reports/42?t=validtoken"
# → HTTP/1.1 302 Found
# → Location: /r/42
# → set-cookie: willbuy_share_42=42:2026-07-23T...:1.abc123...; Path=/r/42; HttpOnly; Secure; SameSite=Lax; Max-Age=7776000
# → cache-control: no-store
# → (empty body)
```

**Path 2 — expired token → 404, no cookie:**
```
curl -v "http://localhost:3000/reports/43?t=expiredtoken"
# → HTTP/1.1 404 Not Found
# → (no set-cookie header)
# → {"error":"report not found"}
```

**Path 5 — valid cookie → 200 + Cache-Control: no-store:**
```
curl -v -H "Cookie: willbuy_share_44=44:2026-07-23T...:1.<valid-hmac>" \
  "http://localhost:3000/reports/44"
# → HTTP/1.1 200 OK
# → cache-control: no-store
# → {"study_id":44,"conv_score":0.75,"paired_delta_json":{"delta":0.1},...}
```

**Path 8 — no token, no cookie, public=true → 200:**
```
curl -v "http://localhost:3000/reports/45"
# → HTTP/1.1 200 OK
# → {"study_id":45,"conv_score":0.75,...}
```

## Test summary

```
✓ apps/api/test/reports.cookie.test.ts (9 tests) 1283ms
[...]

Issue #82, opened 2026-04-25
feat(web+api): domain verification flow — /dashboard/domains/new + /api/domains/verify


NikolayS:
## Sprint 3 — Domain #1: domain verification flow

**Spec:** §2 #1 (target authorization — verified-domain ONLY in v0.1; TXT DNS record `willbuy-verify=<token>` OR `/.well-known/willbuy-verify` HTTP GET OR `<meta name="willbuy-verify">` tag; tokens rotated on verification-list changes; `public_declared` cut from v0.1).

**Depends on:** #79 (session auth), #80 (dashboard shell + route guard).

**Outcome:** An account owner can self-serve verify a domain they own. They visit `/dashboard/domains/new`, enter a domain, choose a verification method (TXT / well-known / meta-tag), see the token to place, click "Verify now" — the server probes all three methods and marks the domain verified if any passes. No manual DB insert required. Verified domains are then eligible for study creation.

## In scope

- **DB migration** `infra/migrations/NNNN_domain_verifications.sql`:
  - `domain_verifications(id, account_id, domain, token, verified_at, method_used, last_checked_at)` table.
  - `account.verified_domains[]` remains the canonical array used by the API gate; verification writes to it on success.
- `apps/api/src/routes/domains.ts`:
  - `POST /api/domains/token` — for a given `domain`, generates or returns existing 22-char `nanoid` token for the account (idempotent); persists to `domain_verifications`.
  - `POST /api/domains/verify` — probes all three methods in parallel (max 5 s timeout each), records `method_used` + `last_checked_at`, on any pass sets `verified_at` and appends `domain` to `accounts.verified_domains`; returns `{ verified: bool, method_used?, error? }`.
  - TXT probe: DNS lookup `_willbuy-verify.<domain>` for `TXT` record containing `willbuy-verify=<token>`.
  - Well-known probe: `GET https://<domain>/.well-known/willbuy-verify` (follow 1 redirect max); body must equal token exactly (trimmed).
  - Meta probe: `GET https://<domain>/` (or `http://` fallback); parse `<meta name="willbuy-verify" content="<token>">` from first 64 KB.
- `apps/web/app/dashboard/domains/new/page.tsx` — stepper UI: (1) enter domain; (2) choose method + show token + instructions; (3) "Verify now" button → calls `/api/domains/verify` → shows success or error with hints.
- Token rotation: if `verified_domains` changes (domain removed or re-verified), issue new token for remaining unverified entries.

## TDD acceptance

[...]

PR #101, opened 2026-04-25
chore(infra): migration-collision guard + next-migration generator (#100)

### LINKED ISSUE #100: infra: pre-commit / CI guard for migration-number collisions
Three migration-number collisions today (Sprint 3): PR #89+#91 on 0014, PR #95+#96 on 0016. Each required a hotfix or in-PR renumber. Add CI guard: `ls infra/migrations/ | grep -oE '^[0-9]{4}_' | sort | uniq -d | wc -l` should be 0. Optional: scripts/next-migration.sh generator for new migrations. Source: PR #98 hotfix + PR #95/#96 renumbers.

NikolayS:
Closes #100.

## Why

Three migration-number collisions in one day (Sprint 3): PR #89+#91 on
`0014`, PR #95+#96 on `0016`, plus a near-miss on PR #82. Each cost a
~30-min hotfix or in-PR renumber. This adds a cheap CI guard and a
generator so the manual "find the highest prefix and +1" step where
parallel PRs collide is gone.

## What

- **`scripts/check-migrations.sh`** — 4 checks, fail-fast:
  1. No duplicate 4-digit prefixes in `infra/migrations/`.
  2. No duplicate 4-digit prefixes in `infra/sqlever/deploy/`.
  3. Every `infra/migrations/00NN_*.sql` has a matching deploy.
  4. Every deploy script has an entry in `infra/sqlever/sqitch.plan`.
- **`scripts/next-migration.sh <slug>`** — picks the next free prefix
  and creates both stubs at once. Reminder to update `sqitch.plan`.
- **`.github/workflows/ci.yml`** — new step `Check migration-number
  collisions (#100)` runs right after `bun install`, before
  typecheck/lint/build/test, so duplicates fail in <30s.
- **`CLAUDE.md`** — one-liner under a new "Working with migrations"
  section pointing engineers at the generator.

## TDD

Two commits — RED (`6fa5c67`) then GREEN (`c2efe9e`).

The test (`tests/check-migrations.test.ts`) builds a temp-fixture copy
of `infra/`, drops a duplicate-prefix file in, and asserts exit 1 +
matching stderr. The real tree is never mutated, so no manual cleanup
is needed.

## Local evidence

Green case (current `main`):

```
$ bash scripts/check-migrations.sh
migration-collision check: OK (18 migrations, all unique, all paired, all in plan)
```

Manually-induced failure (against a temp copy of `infra/`, real tree
untouched):

```
$ tmp=$(mktemp -d) && cp -R infra "$tmp/infra" \
    && cp "$tmp/infra/migrations/0014_share_tokens.sql" \
          "$tmp/infra/migrations/0014_dup.sql" \
    && (cd "$tmp" && bash $PWD/scripts/check-migrations.sh; echo "exit=$?") \
    && rm -rf "$tmp"
ERROR: duplicate migration number prefix(es) in infra/migrations: 0014
0014_dup.sql
0014_share_tokens.sql
exit=1
```

Vitest:

```
$ ./node_modules/.bin/vitest run tests/check-migrations.test.ts
 ✓ |rest| tests/check-migrations.test.ts (2 tests) 83ms
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

## Spec

Operational guard. No spec section affected; touches infra hygiene
only.

## Public-repo audit

No secrets, IPs, or VM identifiers in the diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,13 @@ jobs:
       - name: Install dependencies
         run: bun install --frozen-lockfile
 
+      - name: Check migration-number collisions (#100)
+        # Fail fast on duplicate 4-digit prefixes in infra/migrations/ or
+        # infra/sqlever/deploy/, missing deploy pairs, or missing sqitch.plan
+        # entries. Three collisions in one day (PR #89+#91, PR #95+#96) cost
+        # a ~30-min hotfix each — this is the cheap pre-test guard.
+        run: bash scripts/check-migrations.sh
+
       - name: Cache Playwright browsers
         uses: actions/cache@v4
         with:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -108,6 +108,10 @@ Then:
 
 Use binary units (KiB, MiB, GiB, TiB) for memory, storage, and data sizes in prose, reports, and documentation. Exception: provider-native config formats stay as the provider writes them (e.g. Postgres `shared_buffers = '2GB'`).
 
+## Working with migrations
+
+New migration? Use `bash scripts/next-migration.sh <slug>` — it picks the next free 4-digit prefix and creates both `infra/migrations/` and `infra/sqlever/deploy/` stubs at once. Then add the matching entry to `infra/sqlever/sqitch.plan`. CI runs `bash scripts/check-migrations.sh` (issue #100) to fail fast on prefix collisions.
+
 ## Commands you'll use a lot
 
 ```sh
--- a/scripts/check-migrations.sh
+++ b/scripts/check-migrations.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+IFS=$'\n\t'
+
+# scripts/check-migrations.sh — guard against migration-number collisions.
+#
+# Issue #100. Three collisions in one day (PR #89+#91 on 0014, PR #95+#96 on
+# 0016, plus a near-miss on PR #82) each cost a ~30-min hotfix or in-PR
+# renumber. Run this in CI before tests so duplicates fail fast.
+#
+# Checks:
+#   1. No duplicate 4-digit prefixes in infra/migrations/.
+#   2. No duplicate 4-digit prefixes in infra/sqlever/deploy/.
+#   3. Every infra/migrations/00NN_*.sql has a matching deploy script.
+#   4. Every infra/sqlever/deploy/00NN_*.sql has a sqitch.plan entry.
+#
+# Run path is the cwd, so tests can point it at a fixture tree.
+
+migrations_dir="infra/migrations"
+sqlever_dir="infra/sqlever/deploy"
+plan_file="infra/sqlever/sqitch.plan"
+
+# Check 1: no duplicate prefixes in infra/migrations/.
+dups=$(ls "${migrations_dir}" | grep -oE '^[0-9]{4}' | sort | uniq -d || true)
+if [[ -n "${dups}" ]]; then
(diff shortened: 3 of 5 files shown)


NikolayS:
## VERDICT: APPROVE

(Self-approval blocked — leaving as `--comment` per review protocol.)

**Reviewer:** rev-pr101
**CI:** all green (`build`, `aggregator`, `egress-integration` all `SUCCESS`). New `Check migration-number collisions (#100)` step in `build` printed: `migration-collision check: OK (18 migrations, all unique, all paired, all in plan)`.

### Verified against issue #100 spec

1. **`scripts/check-migrations.sh`** — all 4 checks present and correct:
   - Check 1: duplicate prefixes in `infra/migrations/` → `ls | grep -oE '^[0-9]{4}' | sort | uniq -d`, stderr message lists offending files, exit 1.
   - Check 2: duplicate prefixes in `infra/sqlever/deploy/` → same pattern.
   - Check 3: `comm -23 <(ls migrations) <(ls deploy)` → catches migration-without-deploy.
   - Check 4: every deploy script has `^${name}\b` entry in `sqitch.plan`.
   - Uses `set -Eeuo pipefail` + `IFS=$'\n\t'` (CLAUDE.md shell-style compliant; `-E` is stricter than spec's `-e`, fine).
   - Stdout for OK, stderr for errors. Read-only → idempotent.

2. **`scripts/next-migration.sh`** — picks next 4-digit prefix via `printf "%04d" $((10#${last}+1))` (the `10#` prefix correctly forces base-10 so `0008`/`0009` don't get parsed as octal). Touches both `infra/migrations/` and `infra/sqlever/deploy/` stubs. Prints reminder to update `sqitch.plan` (does not auto-append — sane).

3. **`.github/workflows/ci.yml`** — new step lives between `Install dependencies` and `Cache Playwright browsers`, before typecheck/lint/build/test. Fails fast as advertised.

4. **`CLAUDE.md`** — new "Working with migrations" section is one heading + one paragraph (3 lines). Doesn't bloat the file.

5. **`tests/check-migrations.test.ts`** — uses `mkdtempSync(tmpdir(), 'willbuy-mig-guard-')` + `cpSync` to clone `infra/` into a temp dir; mutates only the fixture; cleans up via `afterEach` with `rmSync(..., { recursive: true, force: true })`. Test 1 (clean repo → exit 0) and Test 2 (induced 0014 duplicate → exit 1, stderr matches `/duplicate migration number prefix/i` and contains `0014`) both present. Test 3 (missing-deploy) was optional — skipped, fine.

6. **TDD discipline:** RED `6fa5c67` is `tests/check-migrations.test.ts` only (73 insertions). GREEN `c2efe9e` adds script + CI + CLAUDE.md (and trims 13 lines of test that were stub-only). No amend, no force-push, no `--no-verify` — verified via `git log` + reflog.

7. **§2 #12 forbidden identifiers** — grep on the new files: none.

### Minor nits (non-blocking, no need to gate merge)

- **N1 (`next-migration.sh`):** if `infra/migrations/` were empty, `((10#${last}+1))` would error under `set -u` because `10#` needs a non-empty operand. Not reachable in willbuy (`0000_init.sql` exists since day 1). One-line defensive fix: `last="${last:-0000}"` before the arithmetic.
- **N2 (`next-migration.sh`):** `touch` silently succeeds on pre-existing files. A `set -C` (noclobber) + `>file` would surface "this prefix already exists" explicitly. Only matters when the tree is already inconsistent, which `check-migrations.sh` would catch on the next CI run anyway.
- **N3 (`check-migrations.sh` check 3):** `comm -23` is one-way (catches "migration without deploy"). A reverse `comm -13` would catch orphan deploy scripts directly; check 4 (`sqitch.plan`) effectively covers them already, so this is purely cosmetic.

LGTM. Ship it after manager approval.

Issue #93, opened 2026-04-25
chore(infra): rev-pr91 nits — dual-maintenance warning + tombstone comment + test fixture


NikolayS:
Non-blocking findings from rev-pr91 review of merged PR #91 (sqlever):

**LOW-1 — Dual-maintenance hazard** (most important):
\`infra/migrations/0001–0013_*.sql\` (historical reference) and \`infra/sqlever/deploy/\` (operative) coexist. Future editors might modify the wrong one and silently drift. **Fix:** add a `README.md` in \`infra/migrations/\` that says: "These files are historical reference only. The operative source is \`infra/sqlever/deploy/\`. Do not modify these — modify the corresponding deploy script instead." OR add a stub `_README.txt` per migration. Or delete the originals entirely (reviewer flagged this as the cleanest path; check whether anything in the codebase still references \`infra/migrations/\` paths via grep before deleting).

**LOW-2 — Tombstone comment misleading.**
\`infra/migrations/0014_seed_sqlever_state.sql\`'s prose claims to "seed sqitch.* state tables" but the deploy script only writes a `_migrations` row (sqlever self-seeds via its own internal mechanism). Update the comment to accurately describe what it does (or doesn't do).

**LOW-3 — Failing-migration test fixture relies on savepoint semantics.**
The test fixture for the rollback case includes \`BEGIN\`/\`COMMIT\` inline; works because of Postgres savepoint behavior, but fragile if sqlever's transaction model changes. Consider rewriting to not rely on savepoints OR add a comment explaining the dependency.

**MEDIUM-1 — Squash commit ate red→green TDD history.**
PR #91 was a single commit (no separate red→green pair) so future bisects against migration test failures will be harder. Acknowledged tradeoff for a sweep-style cutover; not actionable retroactively. File for awareness, no fix.

Source: rev-pr91 review.

PR #108, opened 2026-04-25
chore(infra): rev-pr91 nits — README + tombstone comment + savepoint note (#93)

NikolayS:
Closes the three LOW-severity nits from the rev-pr91 review on issue #93.

## Fixes

- **LOW-1 — Dual-maintenance warning.** New `infra/migrations/README.md`
  marks the dir as historical reference and points future editors at
  `infra/sqlever/deploy/` and `scripts/next-migration.sh`. The CI guard
  in `scripts/check-migrations.sh` is the durable enforcement; this
  README is just for human readers who land in the wrong dir.

- **LOW-2 — Tombstone comment in `0015_seed_sqlever_state.sql`.** The
  old header claimed to "seed sqlever's sqitch.* state tables" — the
  deploy script only writes a single `_migrations` shadow row. Headers
  on **both** copies (`infra/migrations/` historical + `infra/sqlever/deploy/`
  operative) now accurately describe the payload: a single ON CONFLICT
  insert into `_migrations` to keep the row count aligned with the .sql
  file count; sqlever manages its own `sqitch.*` state itself.

- **LOW-3 — Failing-migration fixture savepoint note.** The fixture in
  `tests/migrations.test.ts` embeds inline `BEGIN; … COMMIT;` and
  relies on sqlever's outer-transaction model (the inner BEGIN/COMMIT
  is treated by Postgres as an implicit savepoint inside sqlever's
  wrapping transaction). Added an inline comment near the fixture
  explaining the dependency and what would break if sqlever changes
  its transaction model.

## Drive-by

`scripts/check-migrations.sh` previously did `comm` over the raw `ls`
of both migration dirs, which meant adding the new `infra/migrations/README.md`
broke the guard ("missing in `infra/sqlever/deploy/`"). Filtered the
diff and the migration count to `*.sql` so non-SQL artefacts don't
trip the check. No prefix or pairing logic changed.

## Skipped

- **MEDIUM-1** (squash commit ate red→green TDD history) — not
  actionable retroactively, per #93.

## Verification

- `bash scripts/check-migrations.sh` → `migration-collision check: OK
  (19 migrations, all unique, all paired, all in plan)`
- Docs/comments only otherwise; no test changes that affect runtime.

### DIFF
--- a/infra/migrations/0015_seed_sqlever_state.sql
+++ b/infra/migrations/0015_seed_sqlever_state.sql
@@ -1,23 +1,21 @@
--- 0015_seed_sqlever_state.sql — one-shot state seed for NikolayS/sqlever (issue #48, amendment A4).
+-- 0015_seed_sqlever_state.sql — HISTORICAL REFERENCE (issue #48, amendment A4 + PR #91).
 --
--- Context: PR #46 landed a bash migrate.sh that tracked applied migrations in
--- _migrations(filename, checksum, applied_at). Amendment A4 (2026-04-24)
--- replaces the bash runner with NikolayS/sqlever, which tracks state in the
--- sqitch.* schema (sqitch.changes, sqitch.events, sqitch.projects).
+-- This file is HISTORICAL REFERENCE ONLY. See infra/migrations/README.md.
+-- The operative source for this migration is
+-- `infra/sqlever/deploy/0015_seed_sqlever_state.sql`.
 --
--- For databases that ran the old bash runner (i.e. _migrations is already
--- populated with rows 0000–0013), sqlever would otherwise try to re-apply
--- every migration on the first `sqlever deploy` run.
+-- What this migration actually does (in the operative deploy script):
+-- it seeds the legacy `_migrations` shadow row for 0015 so the row count
+-- stays consistent with the .sql file count in infra/migrations/ (which
+-- the test suite uses as the expected row count).
 --
--- This migration seeds sqlever's sqitch.* state tables with one row per
--- existing _migrations entry so that sqlever sees 0000–0013 as already
--- deployed and only applies genuinely new migrations going forward.
---
--- Idempotency: the INSERT uses ON CONFLICT DO NOTHING, so re-running this
--- migration on a database that already has the sqitch.* rows is a no-op.
--- The sqlever deploy scripts (infra/sqlever/deploy/) use IF NOT EXISTS /
--- ON CONFLICT guards on all schema DDL, so they are also safe to re-run.
+-- It does NOT seed sqlever's own state tables. sqlever manages its own
+-- sqitch.* state (sqitch.changes, sqitch.events, sqitch.projects) on its
+-- own — this migration does not write to those tables. The bash-to-sqlever
+-- transition is handled outside this migration.
 --
 -- NOTE: This file is intentionally kept in infra/migrations/ so that the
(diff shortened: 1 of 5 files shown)


NikolayS:
## VERDICT: APPROVE

rev-pr108 review of LOW-1/LOW-2/LOW-3 nits + drive-by `check-migrations.sh` filter. CI green (aggregator/build/egress-integration all pass).

### Findings: 0 blockers

**1. `infra/migrations/README.md` (LOW-1).** 14 lines, well under the 20-line ceiling. Points at `scripts/next-migration.sh`, calls out the dual-dir pattern, references the PR #101 CI guard, says "Do NOT modify files here directly". Doesn't drift into spec content. Good.

**2. `0015_seed_sqlever_state.sql` headers (LOW-2).** Both copies rewritten correctly:
   - Historical copy (`infra/migrations/`): explicitly marked HISTORICAL REFERENCE, points at the operative script, and the key correction is in: "It does NOT seed sqlever's own state tables. sqlever manages its own sqitch.* state ... on its own — this migration does not write to those tables."
   - Operative copy (`infra/sqlever/deploy/`): clean "What this script DOES / does NOT do" structure. "single ON CONFLICT DO NOTHING insert into _migrations" — accurate. "sqlever manages those itself when it deploys this change" — accurate.
   
   Headers are appropriately differentiated (one is a tombstone pointer, the other is the live header) rather than byte-identical, which is the right call for the dual-source pattern.

**3. `tests/migrations.test.ts` savepoint comment (LOW-3).** Inline at the fixture site (not at top of file). Covers both required points: (a) the inline `BEGIN/COMMIT` is treated by Postgres as an implicit SAVEPOINT inside sqlever's outer transaction, and (b) explicit failure modes if sqlever's transaction model changes ("leak the half-created willbuy_fail_fixture table" / "commit the inner block"). Even includes a debugging hint ("If those assertions below start failing mysteriously, audit sqlever's transaction-wrapping behaviour first"), which is the kind of breadcrumb future-me will appreciate.

**4. `scripts/check-migrations.sh` drive-by.** Filter is minimal and additive:
   - Check 3 `comm -23` inputs both filtered through `grep '\.sql$'` ✓
   - Final `count` line changed to `grep -E '^[0-9].*\.sql$' | wc -l | tr -d ' '` ✓
   - Check 1/2 prefix extraction (`grep -oE '^[0-9]{4}'`) is unchanged — README.md doesn't start with a digit so it can't pollute the dup check; no regression.
   - Check 4 (sqitch.plan) loop already used `[0-9]*.sql` glob — independent, unchanged.
   - Regression scenarios: dup `0014_*.sql` still caught (prefix extraction). Missing-deploy still caught (both sides .sql-filtered). Missing-plan-entry still caught (independent loop). All good.

**5. CI guard line.** PR body confirms `migration-collision check: OK (19 migrations, all unique, all paired, all in plan)`. Count is now strictly `.sql` files, which is what the message implies.

**6. TDD discipline.** Docs/comments + a defensive shell-filter. No new tests required (no runtime behaviour change). Single commit is fine.

**7. Hygiene.** No `--no-verify`, no skipped lint, all CI green.

Engineer was upfront in the PR body about the `check-migrations.sh` scope deviation and explained why it was needed (adding README.md would otherwise trip the guard). Appropriate.

DO NOT MERGE pending self-approval-blocked policy — leaving as comment-only approval.

— rev-pr108 (Claude Opus 4.7 1M)

PR #103, opened 2026-04-25
feat(web+api): domain verification flow (#82)

NikolayS:
Closes #82.

## Summary

Implements the v0.1 domain-verification flow per spec §2 #1. An account owner can enter a domain on `/dashboard/domains/new`, publish any one of three challenge proofs (DNS TXT, `/.well-known/willbuy-verify`, or a `<meta>` tag), click **Verify now**, and have the domain atomically appended to `accounts.verified_domains` so subsequent `POST /studies` calls accept that domain's URLs.

- New table: `domain_verifications(id, account_id, domain, verify_token, verified_at, last_checked_at, created_at)` — UNIQUE `(account_id, domain)`.
- New routes (session-gated):
  - `POST /api/domains` — issue a 22-char nanoid challenge token (upserts; rotates on re-request).
  - `POST /api/domains/:domain/verify` — probes all three methods in parallel with a strict 5s per-probe timeout; on first match marks `verified_at` and atomically appends to `accounts.verified_domains`.
- New page: `/dashboard/domains/new` — single-page form → instructions → verify-now → ✅ Verified, redirects to `/dashboard/domains` (issue #83's route).

## Migration number used

**0018** (0017_studies_urls landed on main first; #82's spec said "0017 OR next free" so picked 0018).

## Notes / deltas vs the issue body

The issue body proposed `POST /api/domains/token` and `POST /api/domains/verify` (with body-carried domain), and a TXT lookup at `_willbuy-verify.<domain>`. The orchestrator-spec for this PR called for `POST /api/domains` + `POST /api/domains/:domain/verify` and a TXT lookup on the bare `<domain>` for `willbuy-verify=<token>`. I implemented the orchestrator spec; the issue body can be reconciled in a follow-up if needed (not breaking — easy to add aliases).

Out of scope (per the spec): background re-verification job (Sprint 4 — manual "Verify now" only), `public_declared` mode (cut from v0.1), IDN/wildcard domains.

## Curl evidence (recorded against a real Postgres-in-Docker instance via `app.inject` at PR time)

```
$ curl -X POST http://localhost/api/domains -b 'wb_session=...' -H 'content-type: application/json' -d '{"domain":"example.com"}'
HTTP/200
[...]

Issue #83, opened 2026-04-25
feat(web+api): domain list page /dashboard/domains — list, re-verify, remove


NikolayS:
## Sprint 3 — Domain #2: domain list page /dashboard/domains

**Spec:** §2 #1 (verified-domain authorization — tokens rotated on verification-list changes), §4.1 (web app: Next.js 14 + Tailwind).

**Depends on:** #82 (domain verification flow).

**Outcome:** An authenticated user can visit `/dashboard/domains` to see all their verified and pending domains, trigger re-verification of any domain, and remove a domain (which also removes it from `accounts.verified_domains` and invalidates any in-flight studies targeting that domain — or blocks removal if a study is in-flight).

## In scope

- `apps/api/src/routes/domains.ts` additions:
  - `GET /api/domains` — returns list `{ domain, verified_at, last_checked_at, method_used }[]` for the authenticated account.
  - `DELETE /api/domains/:domain` — removes from `domain_verifications`; removes from `accounts.verified_domains[]`; blocks if any study with `status IN ('pending','capturing','visiting','aggregating')` targets that domain (returns 409 with explanation).
- `apps/web/app/dashboard/domains/page.tsx`:
  - Table: domain, status badge (Verified / Pending / Failed), verified date, last-checked date, method used.
  - "Re-verify" button → calls `POST /api/domains/verify` inline; updates row status.
  - "Remove" button → confirms ("Any future studies targeting this domain will be rejected") → calls `DELETE /api/domains/:domain`.
  - "Add domain" link → `/dashboard/domains/new`.
  - Empty state: "No domains yet — add your first domain to run a study".
- Token rotation on removal: when a domain is removed from `verified_domains`, any remaining unverified entries for the same account get fresh tokens (matches spec §2 #1 rotation rule).

## TDD acceptance

`apps/api/test/domains.list.test.ts`:

1. `GET /api/domains` with 2 verified domains → returns both with correct fields.
2. `GET /api/domains` with no domains → returns empty array.
3. `DELETE /api/domains/:domain` (no in-flight studies) → 200; domain gone from list and from `accounts.verified_domains`.
4. `DELETE /api/domains/:domain` (study in-flight targeting that domain) → 409 with message.
5. After removal: `POST /studies` with that domain's URL → 422 (unverified).
6. Cross-account: cannot delete another account's domain → 404.

## Out of scope

- Bulk domain import.
- Subdomain wildcard verification.
- Domain-level study listing (show studies per domain).

[...]

PR #107, opened 2026-04-25
feat(web+api): domain list page /dashboard/domains (#83)

NikolayS:
Closes #83.

## Summary

- **API** — adds three handlers in `apps/api/src/routes/domains.ts` (existing POSTs untouched):
  - `GET /api/domains` — list of `{ domain, verify_token, verified_at, last_checked_at, created_at }` for the authenticated account, ordered `created_at DESC`.
  - `DELETE /api/domains/:domain` — atomic transaction: deletes the `domain_verifications` row and removes the value from `accounts.verified_domains[]`. 204 on success; 404 indistinguishably for "doesn't exist" and "owned by another account" (spec §2 #20, no existence leak).
  - `POST /api/domains/:domain/delete` — CSP-safe form-submit fallback for the dashboard list page. Same SQL as DELETE; 302-redirects to `/dashboard/domains`. Required because HTML forms can only emit GET/POST and the dashboard CSP forbids inline JS, so a literal `<button onclick=fetch(...,{method:'DELETE'})>` is not an option.

- **Web** — adds `apps/web/app/dashboard/domains/page.tsx` (Server Component) + `DomainsListView.tsx` (pure renderer, mirrors the `DashboardView` pattern from PR #102):
  - Table: domain, status badge (Verified / Pending), verified_at, last_checked_at, action buttons.
  - Empty state with CTA → `/dashboard/domains/new` (verify flow from PR #103).
  - Top "Add domain" button → `/dashboard/domains/new`.
  - "Re-verify" button → native `<form action="/api/domains/<d>/verify" method="post">` (PR #103 endpoint).
  - "Remove" button → native `<form action="/api/domains/<d>/delete" method="post">` (this PR's form fallback).

- **CSP** — every action is a native HTML form. No inline JS, no `onClick`, no `fetch()` from this page. Works under the strict `/dashboard/*` CSP from PR #80 unchanged.

## Test plan

API (`apps/api/test/domains-list.test.ts`, 8/8 passing):
- [x] AC1 GET ordered DESC
- [x] AC2 GET no session → 401
- [x] AC3 cross-account isolation
- [x] AC4 DELETE removes from both `domain_verifications` and `accounts.verified_domains`
- [x] AC5 DELETE non-existent → 404
- [x] AC6 DELETE other account's domain → 404 (no leak); their row stays intact
- [x] AC7 DELETE no session → 401
- [x] AC8 POST `.../delete` form-fallback → 302 to `/dashboard/domains`

[...]

PR #109, opened 2026-04-25
fix(api): tighten domain-verify probes (#104)

### LINKED ISSUE #104: fix(api): tighten domain-verify probes — HTTPS-only + redirect re-check + NODE_ENV guard on test seam
Five non-blocking findings from rev-pr103 review of merged PR #103. Two are real security concerns worth fixing before launch:

**SEC-1 — HTTPS-only claim is wrong.**
PR #103 body says "All probes use HTTPS only (no plaintext fallback)". Reality: \`fetchWithFallback\` in \`apps/api/src/routes/domains.ts\` tries https then falls back to http for both well-known and meta probes. Active-MITM on the http leg can mint a fake "verified" status.

**Fix:** drop the http fallback. HTTPS only. If a domain doesn't have HTTPS configured, the user must use the DNS TXT method (which doesn't have this concern).

**SEC-2 — Meta probe redirect-follow without eTLD+1 re-check.**
\`probeMeta\` uses \`redirect: 'follow'\` and accepts the verify token from whatever URL it lands on — including a redirect to a different eTLD+1. Narrow attack window but inconsistent with \`probeWellKnown\` which uses \`redirect: 'manual'\`.

**Fix:** either (a) match \`probeWellKnown\`'s \`redirect: 'manual'\` policy, OR (b) re-check eTLD+1 of the final URL against the requested domain before parsing the body.

**N1 — \`__test_setProbes()\` exported without NODE_ENV guard.**
The original review brief required env-gating; sibling PRs in this repo (e.g. \`WILLBUY_DEV_SESSION\`) follow that pattern. Wrap \`__test_setProbes\` in \`if (process.env.NODE_ENV === 'test')\`.

**N2 — TDD: tests + impl landed in same commits.**
Sibling pattern in this repo (PRs #94/#95/#96), so already-established practice. Flagged for transparency, no fix needed.

**N3 — Cross-eTLD+1 redirect policy undocumented.**
Document the chosen policy in a comment near the probe functions (tied to SEC-2 fix).

Source: rev-pr103 review at https://github.com/NikolayS/willbuy/pull/103#issuecomment-4320709721

NikolayS:
Closes #104.

Three security/correctness fixes against the domain verification probes in `apps/api/src/routes/domains.ts` (originally landed in #103). Scope is intentionally tight — only `routes/domains.ts` and `test/domains.test.ts` are touched. The POST `/api/domains` create handler and the #83 list/delete endpoints are not modified.

### SEC-1 — drop the http:// fallback (HTTPS-only)

The pre-fix `fetchWithFallback` tried `https://<domain>/...` first and silently fell back to `http://<domain>/...` for both `probeWellKnown` and `probeMeta`. An active in-path attacker on the http leg could mint a fake "verified" status by returning the published token (or a redirect that follows to it).

**Fix:** removed `fetchWithFallback`; introduced `httpsProbeFetch(url, init)` which:
- requires `url.startsWith('https://')` (defense-in-depth so a future refactor can't reintroduce http);
- runs the per-probe 5s `AbortController` timeout in one place;
- returns `null` on any non-2xx response (centralizes the redirect-rejection policy used by both probes).

Domains without TLS must use the DNS TXT method, which is not vulnerable to in-path HTTP injection in the same way.

### SEC-2 — meta probe rejects redirects (no cross-eTLD+1 token leak)

The pre-fix `probeMeta` used `redirect: 'follow'`, so a 302 from the verified domain to (say) `attacker.example/` would be transparently followed and the attacker's response body scanned for the token. `probeWellKnown` already used `redirect: 'manual'`; the meta probe is now consistent.

**Fix:** `probeMeta` now passes `redirect: 'manual'` to `httpsProbeFetch`; any 3xx response is dropped without reading or scanning the body. Both probe functions carry a comment documenting the policy and the threat model (issue #104 SEC-2).

### N1 — `__test_setProbes()` is NODE_ENV-gated

The pre-fix exports were live in any environment. A misuse (e.g. accidental import-on-startup) could swap in permissive probes that mint fake verifications.

**Fix:** both `__test_setProbes()` and `__test_resetProbes()` early-return when `process.env.NODE_ENV !== 'test'`. The exports remain so test files can import them unconditionally; the functions just no-op outside `test`. Mirrors the `WILLBUY_DEV_SESSION` pattern in `routes/auth.ts`.

## Test outputs

TDD: RED commit (`208d14b`) added three failing ACs; GREEN commit (`27d6113`) makes them pass. All 13 ACs pass under the docker-gated real-DB suite.

```
$ bun --cwd apps/api test domains.test

 RUN  v2.1.9 /Users/nik/github/willbuy-wt-104-domain-sec/apps/api

 ✓ test/domains.test.ts (13 tests) 6327ms
   ✓ domains routes (issue #82, real DB) > AC9: probes time out at 5s and request returns within ~6s 5007ms

 Test Files  1 passed (1)
      Tests  13 passed (13)
```

New tests:
- **AC10 (SEC-1)** — stub https-fetch to throw `ECONNREFUSED`; assert `verified=false` AND every URL passed to the fetch seam starts with `https://` (no `http://` fallback attempted).
- **AC11 (SEC-2)** — stub fetch so it returns a 200-with-token when the route asks for `redirect: 'follow'` (the pre-fix bug surface), but a 302 to `attacker.example` when the route asks for `redirect: 'manual'`. With the fix in place, the route asks for `manual`, sees the 302, and returns `verified=false`. Without the fix, the route asks for `follow`, the mock simulates a successfully-followed redirect with the attacker's body, and `verified=true` would slip through.
- **AC12 (N1)** — flip `process.env.NODE_ENV='production'`, attempt to inject a malicious DNS-positive probe via `__test_setProbes`, restore env, run verify on a non-existent `.example` domain. With the guard, defaults remain; real DNS errors out → `verified=false`. Without the guard, the malicious probe would mint `verified=true`.

Pre-flight clean:
```
$ bun --cwd apps/api typecheck     # tsc -p tsconfig.test.json — clean
$ bun run lint                     # eslint . — clean
```

## Before/after evidence

The route now behaves as follows on adversarial inputs (these are the scenarios encoded in AC10/AC11 — the test seam lets us drive them without any external network):

**Scenario A — domain serves no HTTPS (only http://).**

- *Before:* `httpsProbeFetch('https://...')` rejects with ECONNREFUSED → `fetchWithFallback` retries `http://<domain>/.well-known/willbuy-verify`; if an in-path attacker returns the token, the route returns `{ verified: true, method: 'well_known' }`.
- *After:* `httpsProbeFetch` is the only path; on ECONNREFUSED it returns `null`. The route returns `{ verified: false }` and the user is told (via the spec's error vocabulary) that the well-known/meta probes require HTTPS. Equivalent curl:
  ```
  $ curl -sX POST -b "wb_session=…" \
      https://willbuy.dev/api/domains/no-tls.example/verify
  {"verified":false}
  ```

**Scenario B — root page 302-redirects to attacker.example.**

- *Before:* `probeMeta` ran fetch with `redirect: 'follow'`. node:fetch silently followed the 302 to `attacker.example`, scanned the attacker's body, found `<meta name="willbuy-verify" content="…">`, and returned `{ verified: true, method: 'meta' }`. **Token-leak path closed.**
- *After:* `probeMeta` passes `redirect: 'manual'`; the 302 has `res.ok === false` and `httpsProbeFetch` returns `null` without ever reading the body. The route returns `{ verified: false }`:
  ```
  $ curl -sX POST -b "wb_session=…" \
      https://willbuy.dev/api/domains/redirected.example/verify
  {"verified":false}
  ```

**Scenario C — outside-test code calls `__test_setProbes`.**

- *Before:* the call mutated `_resolveTxt`/`_fetch` and the next probe used the injected impls.
- *After:* `process.env.NODE_ENV !== 'test'` causes an early return; the defaults remain, and a real verify probe runs.

## Test plan

- [x] 10 existing ACs (AC1–AC9 incl. AC3b) still pass under the real-DB docker suite.
- [x] AC10 (SEC-1) green: no `http://` URL is ever passed to the fetch seam.
- [x] AC11 (SEC-2) green: a 302 to a different eTLD+1 yields `verified=false`.
- [x] AC12 (N1) green: `__test_setProbes` is a no-op under `NODE_ENV=production`.
- [x] `bun --cwd apps/api typecheck` clean.
- [x] `bun run lint` clean.
- [ ] Reviewer: confirm scope — only `apps/api/src/routes/domains.ts` and `apps/api/test/domains.test.ts` changed.
- [ ] Reviewer: confirm the create-handler (POST /api/domains) and #83 list/delete handlers are untouched.

Do not merge.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/api/src/routes/domains.ts
+++ b/apps/api/src/routes/domains.ts
@@ -47,13 +47,21 @@
  *     the magic-link token in routes/auth.ts. Cleartext storage is fine here:
  *     the token is meant to be published by the user.
  *   - Probes are stubbable via __test_setProbes() so the integration test
- *     can drive each method without depending on real DNS/HTTP.
- *   - The HTTP probes prefer https://<domain> first, then fall back to
- *     http://<domain> (some users won't have TLS yet during onboarding).
- *     5s timeout applies per-attempt.
- *   - We do NOT follow redirects on /.well-known (the file should be served
- *     directly from the origin). Default fetch() follows redirects; we set
- *     redirect: 'manual' for the well-known probe to detect this.
+ *     can drive each method without depending on real DNS/HTTP. The setter
+ *     is NODE_ENV-gated (no-op outside 'test') — see issue #104 N1 and the
+ *     WILLBUY_DEV_SESSION pattern in routes/auth.ts.
+ *   - HTTPS-only (issue #104 SEC-1): probeWellKnown and probeMeta both use
+ *     https:// exclusively. There is NO http:// fallback. An active
+ *     network attacker on the http leg can mint a fake "verified" status;
+ *     domains without TLS must use the DNS TXT method, which is not
+ *     vulnerable to in-path HTTP injection in the same way.
+ *   - Redirect policy (issue #104 SEC-2): BOTH http probes use
+ *     redirect: 'manual' and treat any 3xx as a probe failure. The
+ *     verification target document — the well-known file or the root
+ *     page's <meta> — MUST be served directly from the requested origin.
+ *     An attacker who can mint a redirect (e.g. via a hijacked path or a
+ *     path-handling bug) must not be allowed to leak a verified status
+ *     by serving the token from a foreign eTLD+1.
  *   - Cross-account constraint: the route only operates on the
  *     (req.account.id, domain) pair, so two different accounts can each
  *     verify the same domain independently. accounts.verified_domains is
@@ -84,15 +92,27 @@ const DEFAULT_FETCH: FetchFn = (url, init) => fetch(url, init);
 let _resolveTxt: ResolveTxtFn = DEFAULT_RESOLVE_TXT;
 let _fetch: FetchFn = DEFAULT_FETCH;
 
+/**
+ * Test-only seam: inject mock resolveTxt / fetch implementations.
+ *
+ * NODE_ENV-gated (issue #104 N1): in any non-'test' environment this is a
+ * no-op. The export remains so test files can import it unconditionally,
+ * but a misuse in production cannot mint a fake "verified" status by
+ * swapping in a permissive probe. Mirrors the WILLBUY_DEV_SESSION pattern
+ * in routes/auth.ts.
+ */
 export function __test_setProbes(opts: {
   resolveTxt?: ResolveTxtFn;
   fetch?: FetchFn;
 }): void {
+  if (process.env.NODE_ENV !== 'test') return;
   if (opts.resolveTxt) _resolveTxt = opts.resolveTxt;
   if (opts.fetch) _fetch = opts.fetch;
 }
 
+/** Test-only seam: restore default probes. NODE_ENV-gated, like setProbes. */
 export function __test_resetProbes(): void {
+  if (process.env.NODE_ENV !== 'test') return;
   _resolveTxt = DEFAULT_RESOLVE_TXT;
   _fetch = DEFAULT_FETCH;
 }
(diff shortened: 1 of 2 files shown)


NikolayS (commented):
## VERDICT: APPROVE

Reviewed by rev-pr109. Verified PR #109 against issue #104 spec (§2 #1).

### Verification

**SEC-1 (HTTPS-only)** — `httpsProbeFetch(url, init)` at lines 199-220 of `apps/api/src/routes/domains.ts`:
- Hard-asserts `url.startsWith('https://')` → returns null on http (defense-in-depth comment notes future-refactor protection).
- AbortController with 5s `PROBE_TIMEOUT_MS` timeout, properly cleared in both branches.
- Returns `null` on non-2xx (`if (res.ok) return res; return null;`) and on caught error.
- `probeWellKnown` (line 222) and `probeMeta` (line 238) both call `httpsProbeFetch` with literal `https://${domain}/...` URLs — no http fallback.
- `fetchWithFallback` removed entirely (no dead code).

**SEC-2 (no redirect-follow)** — both probes pass `{ redirect: 'manual' }`. With manual redirects, a 302 has `res.ok === false`, so `httpsProbeFetch` returns null, and the body is never read or scanned. Comments at both call sites + file header document this.

**N1 (env-gated test seam)** — `__test_setProbes` (line 108) and `__test_resetProbes` (line 115) both early-return on `process.env.NODE_ENV !== 'test'`. JSDoc references the WILLBUY_DEV_SESSION pattern. Exports remain so test code imports unconditionally.

**File-header docs** — lines 49-64 document all three design choices (test-seam gate, HTTPS-only, redirect: 'manual') with issue-#104 SEC-1/SEC-2/N1 anchors.

**Tests (AC10–12)** in `apps/api/test/domains.test.ts`:
- AC10 (line 423): stubs fetch to throw ECONNREFUSED, captures every URL, asserts verified=false AND every captured URL starts with https://, never http://. Exercises the real `httpsProbeFetch` https-guard path.
- AC11 (line 468): stub returns 302 → `attacker.example` for `redirect: 'manual'`, else returns 200 with attacker HTML containing the real token. Asserts verified=false. The two-branch fetch mock is correct: it would mint a verified=true under the buggy 'follow' policy, and verified=false under the fixed 'manual' policy.
- AC12 (line 520): resets probes; flips NODE_ENV='production'; tries to install malicious probes that would mint a TXT match and a 200-with-token; restores NODE_ENV; runs verify without further setup. Defaults remain in effect for a non-existent .example domain → verified=false. Cleanly exercises the env-gate.

**Scope discipline** — `git diff origin/main..origin/fix/104-domain-probe-security --stat` shows ONLY `apps/api/src/routes/domains.ts` (+75/-29 net) and `apps/api/test/domains.test.ts` (+151). POST /api/domains create handler, GET /api/domains list, DELETE /api/domains/:domain (PR #107), POST /api/domains/:domain/delete, and all web pages — UNTOUCHED.

**TDD discipline** — RED `208d14b` then GREEN `27d6113`, matching the engineer's claim.

**No `--no-verify`, no skipped lint, no new forbidden identifiers** in the diff.

**CI: ALL GREEN** (build, aggregator, egress-integration — workflow run 24942756182).

### Findings: 0

Self-approval blocked → comment fallback. Ready to merge.

PR #163, opened 2026-04-26
feat(visitor-worker): add bin.ts process entrypoint (#162)

### LINKED ISSUE #162: feat(visitor-worker): add bin.ts process entrypoint to run visiting-phase polling loop
## Context

PR #161 implemented \`apps/visitor-worker/src/poller.ts\` with \`runVisitorPollingLoop(opts: PollVisitorOpts)\`. The function is complete and unit-tested but there is no process-level entrypoint to run it as a daemon.

## Task

Add \`apps/visitor-worker/src/bin.ts\` that wires the polling loop as a runnable Node.js process.

### Required behavior

- Read \`DATABASE_URL\` from env (required; exit 1 with error message if missing).
- Read \`CAPTURE_STORAGE_PATH\` from env (default \`/tmp/willbuy/captures\`).
- Instantiate \`LocalCliProvider\` from \`@willbuy/llm-adapter\` (no extra config needed — it reads \`WILLBUY_LLM_BIN\` / \`WILLBUY_LLM_MODEL\` from env).
- Wire a minimal \`ObjectStorage\` implementation that reads files from \`CAPTURE_STORAGE_PATH/<key>\` using \`node:fs/promises\`. This is a **read-only** subset: only \`get(key)\` and \`has(key)\` are called by the visitor poller. The \`put\` method can throw \`Error('visitor-worker storage is read-only')\`.
- Call \`runVisitorPollingLoop({ pool, storage, provider, signal })\`.
- Handle \`SIGTERM\` / \`SIGINT\` gracefully: set the AbortController signal and await loop exit, then call \`pool.end()\` and exit 0.
- Write a startup line to stdout: \`[willbuy-visitor-worker] starting ...\`.

### Add a \`build\` script to \`apps/visitor-worker/package.json\`

Mirror the \`capture-broker\` pattern: \`"build": "tsc -p tsconfig.json"\`. Add \`tsconfig.json\` if missing (model it on \`apps/capture-broker/tsconfig.json\`, emitting to \`dist/\`). The build must produce \`dist/bin.js\`.

### TDD requirement

Add a smoke test (unit-level is fine, no real DB / LLM needed): import \`bin.ts\` via dynamic import in a test, start it against a mock pool + mock provider, assert it calls \`runVisitorPollingLoop\` once, abort the signal, confirm it exits cleanly. Alternatively, test the wiring by stubbing \`runVisitorPollingLoop\` from \`./poller.js\` and asserting it is called with correctly-shaped opts.

## Spec ref

Spec §5.11 (visitor worker process lifecycle), §5.1 (worker process entrypoints).

## Definition of done

- \`dist/bin.js\` builds without TS errors.
- \`bun run test\` in \`apps/visitor-worker\` passes.
- \`node dist/bin.js\` exits non-zero immediately when \`DATABASE_URL\` is missing (smoke check).

Closes #159 (partial — dogfood prerequisite)

NikolayS:
## Summary

- Adds `apps/visitor-worker/src/bin.ts` — the visiting-phase process entrypoint (spec §5.1, §5.11)
- Exports `createVisitorWorker(env)` factory that validates `DATABASE_URL`, builds a read-only `ObjectStorage` backed by the local filesystem, creates a `pg.Pool`, and instantiates `LocalCliProvider`
- Process entrypoint wires the factory to `runVisitorPollingLoop`, handles `SIGTERM`/`SIGINT` via `AbortController`, and logs the host portion of `DATABASE_URL` (no credentials) on startup
- Adds `apps/visitor-worker/tsconfig.build.json` (mirrors `apps/capture-broker/tsconfig.build.json`) and `build` script (`tsc -p tsconfig.build.json → dist/`)
- Moves `pg` from `devDependencies` to `dependencies` (runtime dependency)

## Test evidence

```
$ bun run test

 RUN  v2.1.9 /Users/nik/github/willbuy/apps/visitor-worker

 ✓ test/logger.test.ts (2 tests) 3ms
 ✓ test/bin.test.ts (5 tests) 3ms
 ✓ test/logicalRequestKey.test.ts (6 tests) 2ms
 ✓ test/runVisit.transportError.test.ts (1 test) 2ms
 ✓ test/runVisit.schemaFail.test.ts (1 test) 3ms
 ✓ test/runVisit.happyPath.test.ts (1 test) 3ms
 ✓ test/runVisit.schemaRepair.test.ts (1 test) 4ms
 ✓ test/poller.test.ts (5 tests) 7ms
 ✓ test/skeleton.test.ts (1 test) 20ms
 ✓ test/lease-release.test.ts (4 tests) 2105ms

 Test Files  10 passed (10)
      Tests  27 passed (27)
```

## Spec refs

- §5.1 — worker process entrypoints
- §5.11 — visitor worker process lifecycle

## Checklist

- [x] `bun run typecheck` passes
- [x] `bun run build` produces `dist/bin.js`
- [x] All 27 tests pass (5 new in `bin.test.ts`)
- [x] No secrets — `DATABASE_URL` logged as host-only
- [x] `pg` moved to runtime `dependencies`

Closes #162

### DIFF
--- a/apps/visitor-worker/package.json
+++ b/apps/visitor-worker/package.json
@@ -13,6 +13,7 @@
     }
   },
   "scripts": {
+    "build": "tsc -p tsconfig.build.json",
     "test": "vitest run",
     "typecheck": "tsc -p tsconfig.test.json"
   },
@@ -20,6 +21,7 @@
     "@willbuy/llm-adapter": "workspace:*",
     "@willbuy/log": "workspace:*",
     "@willbuy/shared": "workspace:*",
+    "pg": "^8.20.0",
     "pino": "^9.5.0",
     "zod": "^3.23.8"
   },
@@ -27,7 +29,6 @@
     "@types/node": "^22.10.2",
     "@types/pg": "^8.20.0",
     "@willbuy/api": "workspace:*",
-    "pg": "^8.20.0",
     "typescript": "^5.7.2",
     "vitest": "^2.1.8"
   }
--- a/apps/visitor-worker/src/bin.ts
+++ b/apps/visitor-worker/src/bin.ts
@@ -50,7 +50,7 @@ export function createVisitorWorker(env: NodeJS.ProcessEnv): VisitorWorkerCompon
         throw err;
       }
     },
-    async put(_key: string, _body: Buffer, _contentType: string): Promise<void> {
+    async put(): Promise<void> {
       throw new Error('visitor-worker storage is read-only');
     },
   };
--- a/apps/visitor-worker/src/bin.ts
+++ b/apps/visitor-worker/src/bin.ts
@@ -1,0 +1,108 @@
+/**
+ * Visitor Worker — CLI entrypoint.
+ *
+ * Production: started by systemd `ExecStart=/usr/bin/node dist/bin.js`.
+ * Environment is injected from `/etc/willbuy/visitor-worker.env`.
+ *
+ * Reads:
+ *   DATABASE_URL             — required; postgres connection string
+ *   CAPTURE_STORAGE_PATH     — optional; default /tmp/willbuy/captures
+ *   WILLBUY_LLM_BIN          — optional; LLM CLI binary (default: claude)
+ *   WILLBUY_LLM_MODEL        — optional; model identity token
+ */
+
+import { readFile, stat } from 'node:fs/promises';
+import { join } from 'node:path';
+import { Pool } from 'pg';
+import { LocalCliProvider } from '@willbuy/llm-adapter';
+import { runVisitorPollingLoop } from './poller.js';
+import type { ObjectStorage } from './poller.js';
+
+export type VisitorWorkerComponents = {
+  pool: Pool;
+  storage: ObjectStorage;
+  provider: LocalCliProvider;
+};
+
+/**
+ * Validate env and build the wiring components.
+ * Throws if DATABASE_URL is missing.
+ * Exported for unit testing.
+ */
+export function createVisitorWorker(env: NodeJS.ProcessEnv): VisitorWorkerComponents {
+  const dbUrl = env['DATABASE_URL'];
+  if (!dbUrl) {
+    throw new Error('DATABASE_URL is required');
+  }
+
+  const captureBasePath = env['CAPTURE_STORAGE_PATH'] ?? '/tmp/willbuy/captures';
+
+  const storage: ObjectStorage = {
+    async get(key: string): Promise<Buffer> {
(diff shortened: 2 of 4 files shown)


NikolayS:
## VERDICT: APPROVE

**Summary:** `bin.ts` correctly wires `DATABASE_URL` → `pg.Pool` → `runVisitorPollingLoop` with clean shutdown semantics; one non-blocking double-signal race exists but does not affect normal operation.

**Findings:**

- [PASS] `DATABASE_URL` validation — both missing-key and empty-string cases are guarded and tested. Startup log prints only the host portion, never credentials. Correct.

- [PASS] `pg` moved from `devDependencies` → `dependencies` — this was the root cause of the missing runtime dep; the fix is correct.

- [PASS] `createVisitorWorker` is exported for unit testing without spawning a subprocess. The factory is pure (no side effects beyond constructing a `Pool`), making it easy to assert shape without a real DB. The 5 new tests cover: missing env, empty-string env, shape of returned components, read-only `put`, and provider identity — all meaningful.

- [PASS] `ObjectStorage` inline re-declaration in `bin.ts` matches the shape exported from `poller.ts` (`get / put / has`). The `put` stub correctly throws `'visitor-worker storage is read-only'` rather than silently swallowing writes.

- [PASS] `isMain` guard uses `fileURLToPath(import.meta.url) === process.argv[1]` — correct for ESM. Tests import `createVisitorWorker` directly and never trigger the entrypoint block.

- [PASS] `AbortController` / `SIGTERM` / `SIGINT` wiring: `controller.abort()` then `await loopPromise` then `pool.end()`. The `runVisitorPollingLoop` sleep respects the signal and resolves immediately on abort, so graceful drain is correct. Any in-flight `pollVisitorOnce` transaction will complete (or its connection will be released via `finally`) before `pool.end()` drains the pool.

- [NON-BLOCKING] Double-signal race: if `SIGTERM` and `SIGINT` arrive concurrently, `shutdown()` runs twice. `pool.end()` called a second time on an already-ended pool throws an error (pg throws `Error: Called end on pool more than once`), and `process.exit(0)` is harmless but redundant. Fix is a one-liner guard (`let shuttingDown = false`). Not blocking — dual signals before the first `await` completes is unlikely in practice and the process exits anyway.

- [PASS] `tsconfig.build.json` mirrors the capture-broker pattern (`composite: true`, `noEmit: false`, `rootDir: src`, `outDir: dist`). `build` script added to `package.json`. No issues.

**Manual test evidence:** present — PR description includes `bun run test` output showing all 27 tests (5 new in `bin.test.ts`) passing, plus `bun run typecheck` and `bun run build` confirmed.

PR #168, opened 2026-04-26
fix(aggregator): column name mismatches + missing conv_score/share_token_hash (#167)

### LINKED ISSUE #167: fix(aggregator): column name mismatches + missing conv_score/share_token_hash in reports INSERT
### Bug 1 — wrong column names in visit SELECT

`_read_visits` in `apps/aggregator/src/aggregator/main.py` queries:
\`\`\`sql
SELECT id, variant, backstory_id, status, output_json FROM visits
\`\`\`
but the actual column names are `variant_idx` (not `variant`) and `parsed` (not `output_json`).

Fix: rename in the SELECT query.

### Bug 2 — JSONB column parsed by psycopg, then json.loads called on a dict

`parsed` is a `jsonb` column. `psycopg` auto-deserializes JSONB to a Python dict. The aggregator then calls `json.loads(output_json)` on an already-parsed dict, which raises `TypeError`.

Fix: since `parsed` arrives as a dict, no `json.loads` call needed — use it directly.

### Bug 3 — reports INSERT missing required columns

`run_study` inserts:
\`\`\`python
db.execute(conn, "INSERT INTO reports(study_id, paired_delta_json) VALUES (%s, %s)", ...)
\`\`\`

But `reports` has two NOT NULL columns with no default:
- `conv_score numeric NOT NULL` — the average `will_to_buy` score across all visits
- `share_token_hash text NOT NULL UNIQUE` — a random URL-safe token (hex or base64)

The INSERT will always fail with a NOT NULL violation.

Fix: compute `conv_score` as the mean of `output["will_to_buy"]` values across visits (skip None), and generate `share_token_hash = secrets.token_hex(16)`. Include both in the INSERT.

### Bug 4 — paired study logic expects "A"/"B" variant keys, but variant_idx is 0/1

`_build_visits_by_backstory` groups by `v["variant"]` and filters for pairs with both "A" and "B" keys. After fixing Bug 1, `variant_idx` values are `0` and `1` (integers) — not the strings "A"/"B". For `single` studies (1 URL, all visits have `variant_idx=0`) the function returns no pairs, which is correct behavior for a single study.

Fix: change the pair completion check from `"A" in v and "B" in v` to `0 in v and 1 in v` (matching the actual integer values). The output dict key should remain `variant_idx` values.

Note: `paired_delta` in `stats.py` may also need the same key change — check before modifying.

## Also check: `provider_attempts` INSERT

The `_PgLedger.record()` inserts into `provider_attempts(study_id, kind, status, duration_ms)`. Verify this matches the actual table schema before touching it.

## TDD

Add or fix a test in `apps/aggregator/tests/` that:
1. Mocks the DB to return visits with `variant_idx` (int) and `parsed` (dict), confirms the report is built correctly.
2. Asserts `conv_score` is the average `will_to_buy` from all visits.
3. Asserts `share_token_hash` is present in the INSERT.

Read the existing aggregator tests before adding to understand the mock patterns.

## Spec ref
Spec §5.6 (aggregator), §17 (report schema).

Closes #86 (dogfood prerequisite)

NikolayS:
## Summary

- Fix column names in `_read_visits`: `variant` → `variant_idx`, `output_json` → `parsed` (Bug 1)
- Handle both `dict` (psycopg JSONB auto-deserialize) and `str` (sqlite TEXT) for the `parsed` column (Bug 2)
- Add `conv_score` and `share_token_hash` to the `reports` INSERT — both are `NOT NULL` in the real schema (Bug 3)
- Change variant pair keys from `"A"`/`"B"` to `0`/`1` in `_build_visits_by_backstory` and `stats.paired_delta` to match `visits.variant_idx` values (Bug 4)
- Update `test_main_e2e.py` schema, seed data, and assertions; update `test_stats.py` fixtures accordingly

Closes #167

**Spec ref:** §5.6, §17

## Test plan

- [x] `docker run --rm --entrypoint python willbuy-aggregator-test -m pytest tests/ -v` → 13 passed
- [ ] CI green

### DIFF
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -14,6 +14,7 @@ function trusts that the caller (API service) has already acquired the row.
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import os
 import secrets
@@ -191,14 +192,12 @@ def run_study(
         ledger = _PgLedger(conn, study_id)
     clusters = _cluster_with_labels(findings, llm_caller=llm_caller, ledger=ledger)
 
-    payload = {
-        "paired_delta": paired.to_dict(),
-        "clusters": clusters,
-    }
+    payload = paired.to_dict()
 
     conv_score = float(sum(_score_visit(v["output"]) for v in visits) / len(visits)) if visits else 0.0
 
-    share_token_hash = secrets.token_hex(16)
+    raw_token = secrets.token_urlsafe(32)
+    share_token_hash = hashlib.sha256(raw_token.encode()).hexdigest()
 
     db.execute(
         conn,
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -56,6 +56,18 @@ class _PgLedger:
         )
 
 
+class _NoOpLedger:
+    """No-op ledger for the CLI path.
+
+    provider_attempts requires account_id (NOT NULL) which the CLI path
+    doesn't have. The API service wires the real ledger; the CLI aggregator
+    skips ledger writes (issue #178).
+    """
+
+    def record(self, row: dict) -> None:
+        pass
+
+
 def _read_visits(conn: Any, study_id: str) -> list[dict]:
     rows = db.fetchall(
         conn,
@@ -137,6 +149,7 @@ def run_study(
     study_id: str,
     conn: Any,
     llm_caller: LLMCaller,
+    ledger: Any = None,
 ) -> None:
     """Finalize a study: cluster findings, label, compute paired stats, write report.
 
@@ -148,7 +161,8 @@ def run_study(
     paired = paired_delta(paired_input)
 
     findings = _collect_findings(visits)
-    ledger = _PgLedger(conn, study_id)
+    if ledger is None:
+        ledger = _PgLedger(conn, study_id)
     clusters = _cluster_with_labels(findings, llm_caller=llm_caller, ledger=ledger)
 
     payload = {
@@ -206,7 +220,7 @@ def cli(argv: list[str] | None = None) -> int:
 
     conn = _connect_from_env()
     try:
-        run_study(study_id=args.study_id, conn=conn, llm_caller=_stub_llm_caller)
+        run_study(study_id=args.study_id, conn=conn, llm_caller=_stub_llm_caller, ledger=_NoOpLedger())
     finally:
         conn.close()
     return 0
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -144,6 +144,32 @@ def _cluster_with_labels(
     return out
 
 
+# Mirrors scoreVisit() in packages/shared/src/scoring.ts (amendment A1).
+_PAID_TIERS: frozenset[str] = frozenset({"express", "starter", "scale", "enterprise"})
+_NEXT_ACTION_WEIGHTS: dict[str, float] = {
+    "purchase_paid_today": 1.0,
+    "contact_sales": 0.8,
+    "book_demo": 0.8,
+    "start_paid_trial": 0.6,
+    "bookmark_compare_later": 0.0,
+    "start_free_hobby": 0.0,
+    "ask_teammate": 0.2,
+    "leave": 0.0,
+}
+
+
+def _score_visit(output: dict) -> float:
+    """Weighted conversion score for one visit (mirrors scoreVisit() in scoring.ts)."""
+    action = output.get("next_action", "leave")
+    tier_today = output.get("tier_picked_if_buying_today", "none") or "none"
+    tier_considered = output.get("highest_tier_willing_to_consider", "none") or "none"
+    if action == "bookmark_compare_later":
+        return 0.3 if tier_today in _PAID_TIERS else 0.0
+    if action == "start_free_hobby":
+        return 0.2 if tier_considered in _PAID_TIERS else 0.0
+    return _NEXT_ACTION_WEIGHTS.get(action, 0.0)
+
+
 def run_study(
     *,
     study_id: str,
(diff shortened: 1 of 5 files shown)


NikolayS:
## VERDICT: REQUEST CHANGES

**Summary:** Fixes column name mismatches (`variant`→`variant_idx`, `output_json`→`parsed`), JSONB/str dual-path handling, adds `conv_score`/`share_token_hash` to the reports INSERT, ports `_score_visit()` from `scoring.ts`, and changes variant pair keys from `"A"`/`"B"` to `0`/`1` throughout.

---

**Findings:**

- [BLOCKING] **`share_token_hash` stores a raw token, not a hash.** `secrets.token_hex(16)` generates a random 16-byte hex string and stores it directly in `reports.share_token_hash`. The migration comment (`0009_reports.sql`) explicitly says this column holds "SHA-256 of the share token; raw token never stored (only its hash)." The TS `commitReport` in `aggregator-lock.ts` follows the same contract — the caller is expected to pass in a hash, not a raw token. Storing the raw token in the hash column breaks the security model for any future share-link lookup code that compares a provided token's SHA-256 against this column. Fix: generate a raw token with `secrets.token_urlsafe(32)`, then store `hashlib.sha256(raw_token.encode()).hexdigest()` in the column (and either discard the raw token for the CLI path or surface it for the API path).

- [BLOCKING] **`paired_delta_json` embeds clusters; `clusters_json` also holds clusters — structural mismatch with TS `commitReport`.** The PR builds `payload = {"paired_delta": ..., "clusters": ...}` and writes `json.dumps(payload)` into `paired_delta_json`. The TS `commitReport` (and `reports.ts` reader) treats `paired_delta_json` as the paired-stats payload alone and `clusters_json` as a separate column. The Python path now writes cluster data into `paired_delta_json` (embedded in `payload["clusters"]`) AND again into `clusters_json`. Any reader expecting only paired stats in `paired_delta_json` will get extra unexpected keys. Fix: strip `clusters` out of `payload` before dumping to `paired_delta_json`, i.e. `payload = {"paired_delta": paired.to_dict()}` and pass `clusters` only as `clusters_json`.

- [NON-BLOCKING] **`_build_visits_by_backstory` return type annotation is stale.** Declared as `dict[str, dict[str, dict]]` but the inner key is now `int` (`0`/`1`), not `str`. Should be `dict[str, dict[int, dict]]`. No runtime impact (Python doesn't enforce annotations), but it misleads static analysis tools and future readers.

- [PASS] Column renames (`variant_idx`, `parsed`) are correct and consistent end-to-end across `_read_visits`, test schema, and seed data.

- [PASS] JSONB/str dual-path handling (`isinstance(parsed_col, dict)`) is correct for psycopg vs. sqlite.

- [PASS] Variant pair key change from `"A"`/`"B"` to `0`/`1` is applied consistently in `main.py`, `stats.py`, and all test fixtures.

- [PASS] `_score_visit()` logic matches `scoreVisit()` in `scoring.ts` exactly — same weights, same bump conditions for `bookmark_compare_later` (0.3) and `start_free_hobby` (0.2), same fallback to 0.0 for unknown actions.

- [PASS] `_NoOpLedger` correctly addresses the CLI path's missing `account_id` problem. The `ledger=None` default with lazy `_PgLedger` init preserves backward compatibility for callers that don't pass a ledger.

- [PASS] `conv_score` computation (`mean of _score_visit across all visits, 0.0 if empty`) is correct.

- [PASS] New `test_score_visit.py` covers all weight cases, both bump paths, missing-field defaults, and unknown actions. `test_main_e2e.py` correctly asserts `conv_score` is float and `share_token_hash` is non-empty.

**Manual test evidence:** PR body includes `docker run --rm --entrypoint python willbuy-aggregator-test -m pytest tests/ -v → 13 passed`. CI checkbox is unchecked. No curl output or log lines from a live Postgres run.

NikolayS:
Fixed both blocking issues from REV review: (1) share_token_hash now stores SHA-256(token_urlsafe(32)) hex digest; (2) paired_delta_json now stores only paired.to_dict() (no embedded clusters).

NikolayS:
## VERDICT: APPROVE

**Summary:** Both blocking issues from the prior REQUEST CHANGES are fully resolved in commit `d0f2bd8`. No new issues introduced by the fix commit. One pre-existing flaky test is noted below.

**Findings:**

- [PASS] **`share_token_hash` now stores a SHA-256 hex digest (blocking issue resolved).** `raw_token = secrets.token_urlsafe(32)` generates the token, `hashlib.sha256(raw_token.encode()).hexdigest()` produces the hash, and only the hash is stored in the column. Both `hashlib` and `secrets` are imported. Contract with `tokenMatchesHash()` in the TS API route is satisfied.

- [PASS] **`paired_delta_json` now stores only `paired.to_dict()` (blocking issue resolved).** `payload = paired.to_dict()` — no `clusters` wrapper key. Clusters go separately into `clusters_json`. The e2e test assertions are updated to match: it directly asserts top-level `mean_delta`, `paired_t_p`, etc. on the parsed payload, and fetches clusters from `clusters_json` via a second query.

- [PASS] **Fix commit scope is narrow and correct.** Commit `d0f2bd8` touches only `main.py` (replace `secrets.token_hex(16)` with `token_urlsafe(32)` + SHA-256, collapse `payload` to `paired.to_dict()`) and `test_main_e2e.py` (update assertions to flat structure + separate clusters query). No unintended changes.

- [NON-BLOCKING] **`_build_visits_by_backstory` type annotation still stale** (`dict[str, dict[str, dict]]` — inner key is now `int`, should be `dict[str, dict[int, dict]]`). Carry-over from previous review; no runtime impact. Can be cleaned up in a follow-up.

- [NON-BLOCKING] **CI `build` job: one flaky test in `dashboard.test.ts`** (`AC3: invalid cookie HMAC → 401` got 200). The `pull_request`-triggered run (24964876904) shows 1 failure; the `workflow_dispatch` run against the same commit `d0f2bd8` (24965098773) passes all 428 tests with 0 failures. The flaky test is an integration test in `apps/api` that is unrelated to any change in this PR (no `apps/api` files were touched). Pre-existing flakiness, not introduced here.

**Aggregator CI:** green on both runs. All 13 aggregator tests pass.

PR #169, opened 2026-04-26
fix(capture-broker): upsert page_captures on study/side conflict (#166)

### LINKED ISSUE #166: fix(capture-broker): pgCaptureStore INSERT should upsert on (study_id, side) unique conflict
## Background

\`page_captures\` has a unique constraint \`uq_page_captures_study_side UNIQUE (study_id, COALESCE(side, ''))\`. For a \`single\`-kind study, all visits share the same URL, so only one row should exist per study. For a \`paired\`-kind study, two rows exist (side='A', side='B').

## Problem

\`pgCaptureStore.insert()\` (in \`apps/capture-broker/src/captureStore.ts\`) uses a plain INSERT. When the capture-worker processes multiple visits for the same single-URL study, the second and subsequent broker calls fail with a Postgres unique constraint violation. The broker returns \`{ ok: false, error: 'db_failed' }\` and the capture-worker marks those visits as \`failed\`.

Workaround used during dogfood run #86: manually set \`visits.capture_id = 1\` for all visits after the first one succeeded.

## Fix

Change the INSERT to an upsert that returns the existing row id on conflict:

\`\`\`sql
INSERT INTO page_captures
  (study_id, url_hash, a11y_storage_key, screenshot_storage_key,
   host_count, status, breach_reason, captured_at)
VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
ON CONFLICT (study_id, (COALESCE(side, '')))
  DO UPDATE SET
    a11y_storage_key       = EXCLUDED.a11y_storage_key,
    screenshot_storage_key = EXCLUDED.screenshot_storage_key,
    url_hash               = EXCLUDED.url_hash,
    host_count             = EXCLUDED.host_count,
    status                 = EXCLUDED.status,
    breach_reason          = EXCLUDED.breach_reason,
    captured_at            = EXCLUDED.captured_at
RETURNING id
\`\`\`

This means the last capture's artifact wins (desired — most recent capture replaces stale ones for the same study/side).

Note: the \`side\` column is not yet sent by the capture-worker (always NULL). For future paired studies, the broker request schema will need to include \`side\`. That's out of scope here.

## TDD

- Add a unit test to \`apps/capture-broker/test/\`: insert for the same \`(study_id, null side)\` twice → second call succeeds and returns the same \`id\` (updated row, not new row).
- No existing tests should regress.

## Spec ref

Spec §5.13 (capture broker), §2 #16 (artifact storage).

Closes #86 (prerequisite for dogfood re-run)

NikolayS:
## Summary

- Changes `pgCaptureStore.insert()` from a plain `INSERT` to an upsert using `ON CONFLICT (study_id, (COALESCE(side, ''::text))) DO UPDATE SET ...`
- Prevents unique constraint violations on `uq_page_captures_study_side` when multiple visits share the same `study_id` + NULL `side` (single-URL studies)
- Adds `test/captureStore.test.ts` with 6 unit tests: validates upsert SQL presence, correct conflict target expression, and the core regression scenario (two inserts with the same `study_id` returning the same `id`)

Closes #166

## Spec ref

§5.13

## Test plan

- [ ] `bun run test` in `apps/capture-broker` — all 50 tests pass (6 new)
- [ ] CI green

### DIFF
--- a/apps/capture-broker/src/captureStore.ts
+++ b/apps/capture-broker/src/captureStore.ts
@@ -74,6 +74,15 @@ export function pgCaptureStore(pool: Pool): CaptureStore {
            (study_id, url_hash, a11y_storage_key, screenshot_storage_key,
             host_count, status, breach_reason, captured_at)
          VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+         ON CONFLICT (study_id, (COALESCE(side, ''::text)))
+           DO UPDATE SET
+             a11y_storage_key       = EXCLUDED.a11y_storage_key,
+             screenshot_storage_key = EXCLUDED.screenshot_storage_key,
+             url_hash               = EXCLUDED.url_hash,
+             host_count             = EXCLUDED.host_count,
+             status                 = EXCLUDED.status,
+             breach_reason          = EXCLUDED.breach_reason,
+             captured_at            = EXCLUDED.captured_at
          RETURNING id`,
         [
           row.study_id,
--- a/apps/capture-broker/src/captureStore.ts
+++ b/apps/capture-broker/src/captureStore.ts
@@ -22,6 +22,8 @@ export type PageCaptureRow = {
   study_id?: number;
   /** Salted SHA-256 of the captured URL (spec §5.12); required for pgCaptureStore. */
   url_hash?: string;
+  /** A/B study side ('A' or 'B'); NULL for single-URL studies. */
+  side?: 'A' | 'B' | null;
 };
 
 export type CaptureInsertResult = {
@@ -71,10 +73,10 @@ export function pgCaptureStore(pool: Pool): CaptureStore {
 
       const result = await pool.query<{ id: string }>(
         `INSERT INTO page_captures
-           (study_id, url_hash, a11y_storage_key, screenshot_storage_key,
+           (study_id, side, url_hash, a11y_storage_key, screenshot_storage_key,
             host_count, status, breach_reason, captured_at)
-         VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
-         ON CONFLICT (study_id, (COALESCE(side, ''::text)))
+         VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
+         ON CONFLICT (study_id, (COALESCE(side, '')))
            DO UPDATE SET
              a11y_storage_key       = EXCLUDED.a11y_storage_key,
              screenshot_storage_key = EXCLUDED.screenshot_storage_key,
@@ -86,6 +88,7 @@ export function pgCaptureStore(pool: Pool): CaptureStore {
          RETURNING id`,
         [
           row.study_id,
+          row.side ?? null,
           row.url_hash,
           row.a11y_object_key,
           row.screenshot_object_key,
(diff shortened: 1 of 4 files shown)


NikolayS:
## VERDICT: REQUEST CHANGES

**Summary:** The upsert conflict target uses `''::text` which likely won't match the index definition, and `side` is missing from the INSERT column list — meaning paired-study captures silently overwrite each other.

**Findings:**

- [BLOCKING]: **Conflict target expression mismatch.** The unique index is defined as `coalesce(side, '')` (no type cast), but the SQL uses `ON CONFLICT (study_id, (COALESCE(side, ''::text)))` with an explicit `::text` cast. PostgreSQL requires the conflict target to match the index expression *exactly* as it appears in `pg_index`. This mismatch will produce a runtime error: `there is no unique or exclusion constraint matching the ON CONFLICT specification`. Fix: drop the `::text` cast to match the index — `ON CONFLICT (study_id, (COALESCE(side, '')))`.

- [BLOCKING]: **`side` column absent from INSERT column list — silent data corruption for paired studies.** The INSERT enumerates `(study_id, url_hash, a11y_storage_key, screenshot_storage_key, host_count, status, breach_reason, captured_at)` — `side` is not included, so every inserted row gets `side = NULL` regardless of whether the study is paired (A/B). For single-URL studies (side IS NULL) the upsert fires correctly. For paired studies, both the A-side and B-side captures land with `side = NULL`, so the second insert upserts over the first rather than inserting a new row. The fix for issue #166 needs `side` added to the column list and a corresponding bind parameter.

- [NON-BLOCKING]: The test "two inserts with same study_id return same id (upsert, no unique violation)" mocks `pool.query` — it validates the SQL string but never exercises the actual constraint path. The assertion that "no unique violation" occurs is vacuously true for a mock. Consider a note in the test name to make clear it is SQL-shape only, or add a comment. Not blocking.

**Manual test evidence:** absent (test plan checkbox unchecked; no CI link in PR body)

NikolayS:
Fixed two blocking REV findings: (1) removed ::text cast from ON CONFLICT COALESCE expression to match index definition; (2) added side to INSERT column list to prevent silent overwrite of A-variant by B-variant.

NikolayS:
## VERDICT: APPROVE

**Summary:** Both blocking issues from the previous review are resolved. The fix commit correctly drops the `::text` cast and adds `side` to the INSERT column list. CI is green. No new issues introduced.

**Findings:**

- [PASS]: **Conflict target cast removed.** `ON CONFLICT (study_id, (COALESCE(side, '')))` — no `::text` cast, matches the index definition exactly.

- [PASS]: **`side` added to INSERT column list.** Column list is now `(study_id, side, url_hash, ...)` with VALUES bumped from 8 to 9 params and `row.side ?? null` bound as `$2`. Paired-study captures will no longer silently overwrite each other.

- [PASS]: **Tests cover both fixes.** `captureStore.test.ts` asserts (a) `sql` does not contain `''::text` and (b) the column list matches `/study_id\s*,\s*side\s*,/` with `'A'` present in params.

- [NON-BLOCKING]: **PR body still references `''::text` cast** in the Summary description — cosmetic only, the code is correct.

- [NON-BLOCKING]: **Prior non-blocking observation stands** — the "no unique violation" test exercises SQL shape only against a mock, not the actual constraint path. Acceptable for a unit test; the comment is just informational.

Issue #170, opened 2026-04-26
feat(aggregator+web): produce full Report §5.18 schema and wire fetchReport to API


NikolayS:
## Context

The dogfood run (#86) showed that the pipeline can reach \`status='ready'\`, but the web app's report page at \`/r/:slug\` cannot render anything because:

1. **\`fetchReport\` is a stub** — \`apps/web/app/r/[slug]/fetchReport.ts\` always returns \`null\` except for a test fixture. It needs to call \`GET /api/reports/:slug\`.

2. **Aggregator doesn't produce the full Report schema** — \`packages/shared/src/report.ts\` defines \`ReportT\` with seven visualization elements (§5.18 #1–7): \`headline\`, \`paired_dots\`, \`histograms\`, \`next_actions\`, \`tier_picked\`, \`theme_board\`, \`personas\`. The current aggregator only writes \`paired_delta_json\` + \`clusters_json\` + \`conv_score\` to the \`reports\` table — it does not compute the visualization-ready schema.

3. **API response shape doesn't match the Report schema** — \`GET /reports/:slug\` returns \`{study_id, conv_score, paired_delta_json, clusters_json, ...}\` not the \`ReportT\` shape. The web app calls \`Report.parse(result)\` which would fail on any real data.

### A. Extend aggregator to write full \`reports\` blob (§5.18)

Add a \`reports_json\` (or update \`paired_delta_json\`) column that stores the complete \`ReportT\`-shaped JSON:
- \`meta\`: slug = share_token_hash, low_power = (n_ok_visits < 20)
- \`headline\`: from existing \`paired_delta\` computation
- \`paired_dots\`: per backstory, A+B scores, swing direction
- \`histograms\`: 0–10 bin counts per variant
- \`next_actions\`: next_action counts per variant  
- \`tier_picked\`: tier counts per variant
- \`theme_board\`: clusters mapped to {blockers, objections, confusions, questions}
- \`personas\`: per-backstory card data

For single studies (no B variant), all \`*_b\` fields are \`null\`/empty and variant always = 'A'.

### B. Wire \`fetchReport\` to the API

Replace the stub with a real call to \`GET /api/reports/:slug\`. The API server URL comes from \`process.env.WILLBUY_API_URL\` (or Next.js's internal routing if colocated). Handle the public report path (no token needed if \`reports.public = true\`).

### C. Fix API response shape

[...]

PR #171, opened 2026-04-26
feat(shared+api+web): add report_json passthrough for §5.18 display (#170)

NikolayS:
## Summary

Pre-aggregator wiring for issue #170. Sets up the full pipeline so that once the aggregator writes a `report_json` JSONB blob, the API serves it and the web page renders it. Aggregator changes come after #168 merges.

- **`packages/shared/src/report.ts`** — relaxed `histograms`, `next_actions`, and `tier_picked` from `.length(2)` to `.min(1).max(2)`; made `persona.score_b` and `persona.verdict_b` nullable (single-variant studies have no B data)
- **`infra/migrations/0020_report_json.sql`** — new migration adds `report_json JSONB` column to `reports` (NULL until aggregator writes it)
- **`infra/sqlever/deploy/0020_report_json.sql`** — sqlever BEGIN/COMMIT deploy wrapper for the same migration
- **`apps/api/src/routes/reports.ts`** — added `report_json` to `ReportRow` interface, the `SELECT` column list, and the `send200` response body
- **`apps/web/app/r/[slug]/fetchReport.ts`** — replaced stub with real API fetch: forwards `wb_rt_<slug>` cookie, returns `body.report_json ?? null` on 200, null on any non-200; fixture path preserved for dev/preview

## Test plan

- [x] `packages/shared` tests pass (23 tests)
- [x] `apps/api` tests pass (133 tests, all 17 files)
- [x] `apps/web` typecheck passes (no TS errors)
- [ ] CI green

### DIFF
--- a/apps/web/app/r/[slug]/fetchReport.ts
+++ b/apps/web/app/r/[slug]/fetchReport.ts
@@ -1,11 +1,18 @@
 // Server-only seam — fetches the §5.18 report_json blob for a given slug.
 // Fixture path (WILLBUY_REPORT_FIXTURE=enabled + slug=test-fixture) stays
 // active for dev/preview; all other slugs hit the real API.
+//
+// Return value is a discriminated union:
+//   - 'not_found'  API returned 404 (study_id invalid, report expired, or fetch error)
+//   - 'pending'    API returned 200 but report_json is null (aggregator not yet run)
+//   - unknown      API returned 200 with valid report_json (parsed payload)
 
 import { cookies } from 'next/headers';
 import fixture from '../../../test/fixtures/report.fixture.json';
 
 const FIXTURE_SLUG = 'test-fixture';
+
+export type FetchReportResult = 'not_found' | 'pending' | unknown;
 
 function apiBaseUrl(): string {
   const explicit = process.env['WILLBUY_API_URL'] ?? process.env['NEXT_PUBLIC_API_URL'];
@@ -13,7 +20,7 @@ function apiBaseUrl(): string {
   return 'http://localhost:3001';
 }
 
-export async function fetchReport(slug: string): Promise<unknown | null> {
+export async function fetchReport(slug: string): Promise<FetchReportResult> {
   if (
     process.env.WILLBUY_REPORT_FIXTURE === 'enabled' &&
     slug === FIXTURE_SLUG
@@ -40,11 +47,14 @@ export async function fetchReport(slug: string): Promise<unknown | null> {
       headers,
     });
   } catch {
-    return null;
+    return 'not_found';
   }
 
-  if (!res.ok) return null;
+  if (res.status === 404) return 'not_found';
+  if (!res.ok) return 'not_found';
 
   const body = (await res.json()) as Record<string, unknown>;
-  return body['report_json'] ?? null;
+  const reportJson = body['report_json'];
+  if (reportJson === null || reportJson === undefined) return 'pending';
+  return reportJson;
 }
--- a/apps/web/app/r/[slug]/page.tsx
+++ b/apps/web/app/r/[slug]/page.tsx
@@ -20,13 +20,27 @@ export default async function ReportPage({
   const { slug } = await params;
   const result = await fetchReport(slug);
 
-  if (result === null) {
+  if (result === 'not_found' || result === null) {
     return (
       <>
         <main className="mx-auto max-w-3xl px-6 py-16">
           <h1 className="text-3xl font-bold tracking-tight">Report not found</h1>
           <p className="mt-6 text-gray-700">
             No report exists for <code>{slug}</code>, or the share link has been revoked.
+          </p>
+        </main>
+        <ReportCtaBar />
+      </>
+    );
+  }
+
+  if (result === 'pending') {
+    return (
+      <>
+        <main className="mx-auto max-w-3xl px-6 py-16">
+          <h1 className="text-3xl font-bold tracking-tight">Report is being prepared</h1>
+          <p className="mt-6 text-gray-700">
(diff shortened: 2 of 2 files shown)


NikolayS:
## VERDICT: REQUEST CHANGES

**Summary:** One blocking issue in `fetchReport.ts` — the new real-fetch path returns `body['report_json']` raw (unparsed `unknown`), but the `page.tsx` call-site passes that directly into `Report.parse()`; this is fine structurally, but a network error that causes `res.json()` to throw (malformed JSON from a gateway, for example) is silently swallowed by the outer `try/catch`, returning `null` — which is acceptable. However, a more subtle issue exists: when `report_json` is `null` in the DB (aggregator hasn't run yet), `body['report_json'] ?? null` returns `null`, the page renders the "Report not found" message, and the user gets a confusing 404-style screen for a report that does exist but isn't ready. There is no `202 Accepted` / "report is processing" state distinguished from a genuine 404.

**Findings:**

- [BLOCKING]: **`fetchReport.ts` conflates "report not found" with "report exists but `report_json` is null (aggregator not yet run)"**. The API returns HTTP 200 with `report_json: null` when the aggregator hasn't written the blob yet, and `fetchReport` returns `null`, which `page.tsx` renders as "Report not found". A user who lands on their report URL immediately after study completion will see a false 404 with no indication the data is being processed. At minimum the API should return a distinguishable shape (e.g. `{ ready: false }`) or a `202`, and `fetchReport` / the page should handle it separately. As shipped, this is a user-visible correctness bug for the pre-aggregator window.

- [NON-BLOCKING]: **`PersonaCard` swing label renders `0` for single-variant personas.** The fix `(persona.score_b ?? persona.score_a) - persona.score_a` correctly produces `0` when `score_b` is null, so the card shows `+0` with neutral gray — accurate and not a crash, but the swing badge is meaningless noise for single-variant cards. A `score_b !== null` guard on the badge render (similar to what was done for the B score row) would be cleaner. Not blocking.

- [NON-BLOCKING]: **`PersonaDrawer` renders `verdict_b: null` as an empty `<p>`** when `score_b` is non-null but `verdict_b` happens to be null. The Zod schema allows `score_b` non-null + `verdict_b` null (both are independently nullable). The null guard only wraps the whole drawer block on `score_b !== null`, so if an aggregator bug writes a score without a verdict, React renders `null` as empty text — not a crash, but worth noting.

- [PASS]: Schema relaxation (`.length(2)` → `.min(1).max(2)`) is correct and backward-compatible. Existing two-variant fixtures still pass; single-variant data is now accepted.

- [PASS]: Migration files (`0020_report_json.sql`, sqlever deploy wrapper, sqitch.plan) are consistent and use `ADD COLUMN IF NOT EXISTS` — safe for re-run.

- [PASS]: Cookie forwarding in `fetchReport.ts` matches the API's `wb_rt_${slug}` cookie name and the `Path=/r/${slug}` scope (SSR fetch is server-to-server so path scoping doesn't apply — the manual header forwarding is the right approach).

- [PASS]: `Report.parse(result)` is called at the page boundary after `fetchReport` returns — Zod-at-the-boundary rule upheld.

- [PASS]: Test added for the single-variant path; covers schema parse + render without crash.

**Manual test evidence:** absent (CI checkbox unchecked in PR description).

NikolayS:
Fixed blocking REV finding: fetchReport now returns 'pending' when API returns 200 with report_json=null (aggregator not yet run) vs 'not_found' for genuine 404. Page renders 'Report is being prepared' for the pending case.

NikolayS:
## VERDICT: APPROVE

**Summary:** The blocking issue from the previous review is resolved. `fetchReport` now returns a proper discriminated union (`'not_found' | 'pending' | unknown`) and `page.tsx` handles all three states correctly. No new blocking issues introduced.

**Findings:**

- [PASS]: `fetchReport` discriminated union — `'not_found'` is returned on fetch error, 404, or non-2xx; `'pending'` is returned when `report_json` is null/undefined; the parsed `reportJson` payload is returned on a valid 200. All three arms are correct.

- [PASS]: `page.tsx` state handling — `'not_found'` (plus the legacy `null` guard) renders "not found", `'pending'` renders "Report is being prepared / still running", and the payload falls through to the existing Zod-parse + render path. All three states covered.

- [PASS]: `PersonaGrid.tsx` null-safety — `score_b ?? score_a` collapses single-variant swing to 0 (correct); the B-score span and drawer verdict block are conditionally rendered only when `score_b !== null`. Sort comparator updated consistently.

- [PASS]: Regression test — `report-viz.test.tsx` adds a single-variant render test (`score_b=null`, 1-item histogram/next_actions/tier_picked arrays) that guards the nullable path end-to-end.

- [PASS]: Zod schema relaxation — `.length(2)` → `.min(1).max(2)` on `histograms`, `next_actions`, `tier_picked`; `score_b`/`verdict_b` made `.nullable()`. Correct for single-variant studies.

- [PASS]: API passthrough — `report_json` added to `ReportRow`, the `SELECT` query, and the `send200` payload. Migration adds `report_json JSONB` with `IF NOT EXISTS` guard.

- [NON-BLOCKING]: `result === null` guard in `page.tsx` (line `result === 'not_found' || result === null`) is dead code — `fetchReport` no longer returns `null`. Safe to remove in a follow-up cleanup.

- [NON-BLOCKING]: `FetchReportResult = 'not_found' | 'pending' | unknown` — `unknown` is the top type and subsumes the whole union, so TypeScript won't enforce exhaustiveness on the `'not_found'`/`'pending'` arms at the type level. Runtime behaviour is correct. A tagged-union shape (e.g. `{ status: 'ok'; payload: unknown }`) would be stricter, but this matches the spec comment style and is acceptable for now.

PR #172, opened 2026-04-26
feat(aggregator): compute full §5.18 report_json blob in run_study() (#170)

NikolayS:
## Summary

- Adds `_read_backstories(conn, ids)` helper — fetches `backstories.payload` by id list; gracefully returns `{}` when table is absent (sqlite test path)
- Adds `_build_report_json(...)` helper — computes the full §5.18 visualization blob: `meta`, `headline`, `paired_dots`, `histograms` (11-bin per variant), `next_actions`, `tier_picked`, `theme_board` (4 keys from cluster kinds), `personas` (one per backstory with score_a/score_b/verdict_a/verdict_b)
- `run_study()` wires backstory reading + report_json construction and writes it as the 6th column in `INSERT INTO reports`
- `test_main_e2e.py`: adds `report_json TEXT` to the SQLite schema, asserts non-NULL, `meta.slug == share_token_hash`, `histograms` has ≥1 entry, `theme_board` has all four keys

## Dependencies

- **Depends on #168** (`fix/aggregator-column-names-167`) for the `conv_score`, `share_token_hash`, `clusters_json` columns and the `variant_idx`/`parsed` rename — base branch is set accordingly
- **Depends on #171** (`feat/report-json-170`) for the `report_json JSONB` column migration and relaxed shared `Report` Zod schema — must merge before this lands to main

## Test plan

- [x] All 13 aggregator unit + e2e tests pass inside `willbuy-aggregator:dev` Docker image
- [ ] CI green on this PR
- [ ] #168 and #171 merge first before this targets main

### DIFF
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -45,6 +45,8 @@ VALID_NEXT_ACTIONS: list[str] = [
     "start_paid_trial",
     "bookmark_compare_later",
     "start_free_hobby",
+    "ask_teammate",
+    "leave",
 ]
 
 VALID_TIERS: list[str] = ["none", "hobby", "express", "starter", "scale", "enterprise"]
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -317,19 +317,24 @@ def _build_report_json(
             continue
         seen_bs.add(bs_id)
         bs_payload = _bs_map_get(backstory_map, bs_id)
-        pair = visits_by_backstory.get(bs_id, {})
-        a_out = pair.get(0, {})
-        b_out = pair.get(1, {})
         verdict_a = ""
         verdict_b = None
+        score_a_raw: float | None = None
+        score_b_raw: float | None = None
         for vv in visits:
             if vv["backstory_id"] != bs_id:
                 continue
             out = vv["output"]
-            if vv["variant"] == 0 and not verdict_a:
-                verdict_a = (out.get("first_impression") or out.get("reasoning") or "")[:400]
-            if vv["variant"] == 1 and verdict_b is None:
-                verdict_b = (out.get("first_impression") or out.get("reasoning") or "")[:400]
+            if vv["variant"] == 0:
+                if not verdict_a:
+                    verdict_a = (out.get("first_impression") or out.get("reasoning") or "")[:400]
+                if score_a_raw is None and out.get("will_to_buy") is not None:
+                    score_a_raw = float(out["will_to_buy"])
+            elif vv["variant"] == 1:
+                if verdict_b is None:
+                    verdict_b = (out.get("first_impression") or out.get("reasoning") or "")[:400]
+                if score_b_raw is None and out.get("will_to_buy") is not None:
+                    score_b_raw = float(out["will_to_buy"])
         personas.append({
             "backstory_id": str(bs_id),
             "backstory_name": bs_payload.get("name", str(bs_id)),
@@ -339,8 +344,8 @@ def _build_report_json(
             "stack": str(bs_payload.get("managed_postgres", "")),
             "current_pain": str(bs_payload.get("current_pain", "")),
             "entry_point": str(bs_payload.get("entry_point", "")),
-            "score_a": float(a_out.get("score") or 0),
-            "score_b": float(b_out["score"]) if b_out.get("score") is not None else None,
+            "score_a": score_a_raw if score_a_raw is not None else 0.0,
+            "score_b": score_b_raw,
             "verdict_a": verdict_a,
             "verdict_b": verdict_b,
         })
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -194,6 +194,32 @@ def _cluster_with_labels(
     return out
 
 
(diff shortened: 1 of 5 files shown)


NikolayS:
## VERDICT: REQUEST CHANGES

**Summary:** Two blocking fixes from PR #168 are absent in this branch, and `tier_picked` reads the wrong output field name.

**Findings:**

- [BLOCKING]: **`share_token_hash` stores raw token, not SHA-256 digest.** `run_study()` line 434 sets `share_token_hash = secrets.token_hex(16)` and writes that directly to the DB column. PR #168 required this to store a SHA-256 hex digest of a separately-generated secret token (so the raw token can be issued to the user while only the hash is persisted). Fix: `raw_token = secrets.token_urlsafe(32); share_token_hash = hashlib.sha256(raw_token.encode()).hexdigest()`. `hashlib` is not even imported in this branch.

- [BLOCKING]: **`paired_delta_json` still embeds a wrapper dict with `clusters` key.** `payload = {"paired_delta": paired.to_dict(), "clusters": clusters}` is written as `paired_delta_json`. PR #168 required this column to store only `paired.to_dict()` (flat struct, no wrapping key, no embedded clusters). The clusters already have their own `clusters_json` column. The e2e test added in this PR checks `rj["meta"]["slug"]` but never asserts the shape of `paired_delta_json`, so this regression goes undetected.

- [BLOCKING (from PR #168)]: The two issues above are specifically the fixes that PR #168 was supposed to land. They are confirmed absent on branch `feat/aggregator-report-json-170` (verified via `git show origin/feat/aggregator-report-json-170:apps/aggregator/src/aggregator/main.py`). They must be present before this PR merges.

- [NON-BLOCKING]: **`tier_picked` section reads `v["output"].get("tier_picked", "none")` but the VisitorOutput field is `tier_picked_if_buying_today`.** `_score_visit()` reads the right field name. The tier_picked histogram will silently bucket every visit as "none". Fix: change to `v["output"].get("tier_picked_if_buying_today", "none")`.

- [PASS]: `_score_visit()` correctly mirrors `scoreVisit()` in scoring.ts — weights, tier bump for bookmark/free-hobby, null-safety for missing fields. Unit tests in `test_score_visit.py` are complete and correct.

- [PASS]: `_read_backstories()` exception swallow for sqlite / absent table is appropriate; the `int(bs_id)` cast in `_bs_map_get()` correctly tolerates string fixture IDs.

- [PASS]: `_build_report_json()` histogram bin clamping (`max(0, min(10, int(s)))`) and the O(n²) personas loop are both fine for expected panel sizes.

- [PASS]: `_NoOpLedger` is a clean fix for the CLI path; `ledger=None` default preserves the API service path.

**Manual test evidence:** absent (CI checkbox unchecked in PR description)

NikolayS:
Fixed three blocking REV findings: (1) share_token_hash now stores SHA-256(token_urlsafe(32)); (2) paired_delta_json stores only paired.to_dict(); (3) tier_picked lookup uses correct field name tier_picked_if_buying_today.

NikolayS:
## VERDICT: APPROVE

**Summary:** All three blocking issues from the previous review are correctly resolved. No new blocking issues found in the fix commit.

**Findings:**

- [PASS]: **`share_token_hash` stores SHA-256 hex digest.** `hashlib` is now imported. `raw_token = secrets.token_urlsafe(32)` generates the secret; `share_token_hash = hashlib.sha256(raw_token.encode()).hexdigest()` stores the hash. Correct.

- [PASS]: **`paired_delta_json` stores only `paired.to_dict()`.** `payload = paired.to_dict()` with no wrapper dict; clusters go to the separate `clusters_json` column. The e2e test now asserts `"mean_delta" in payload` directly (not `"paired_delta" in payload`), confirming the flat struct. Correct.

- [PASS]: **`tier_picked` uses `tier_picked_if_buying_today` field name.** Both the `tier_picked` histogram loop (`v["output"].get("tier_picked_if_buying_today", "none")`) and `_score_visit()` read the correct field. Correct.

- [NON-BLOCKING]: **`FetchReportResult = 'not_found' | 'pending' | unknown` collapses to `unknown`.** In TypeScript `unknown` is a top type; the union is degenerate and provides no type narrowing after the string equality guards in `page.tsx`. Functionally harmless — Zod's `Report.parse()` at the boundary provides the actual runtime safety — but the type annotation is misleading. Consider `'not_found' | 'pending' | Record<string, unknown>` or a tagged wrapper.

- [NON-BLOCKING]: **`fetchReport.ts` makes unauthenticated API calls.** The comment acknowledges §5.12 cookie auth "lands when the report-API issue ships", so this is a known pre-existing stub. The non-404 → `not_found` silent swallow of 5xx errors is consistent with the intent to avoid leaking internals. Acceptable for current state.

- [NON-BLOCKING]: **E2e test does not verify `share_token_hash` is a valid 64-char hex string.** `assert isinstance(report_row[2], str) and len(report_row[2]) > 0` would pass even if the old `secrets.token_hex(16)` code were restored. A stronger assertion: `assert len(report_row[2]) == 64 and all(c in '0123456789abcdef' for c in report_row[2])`.

- [PASS]: `_score_visit()` unit tests in `test_score_visit.py` are complete and correct — all 8 action variants covered, tier-bump edge cases for both bookmark and free-hobby branches, missing-field defaults, unknown action.

- [PASS]: `_NoOpLedger`, `_read_backstories()`, `_build_report_json()`, and the `ledger=None` default all look clean. No regressions observed.

Issue #20, opened 2026-04-25
chore(llm-adapter): extend AST FORBIDDEN list to 9 identifiers per spec §2 #15


NikolayS:
Reviewer on PR #19 caught a spec/CLAUDE.md/implementation drift:

- **Spec §2 #15 line 36** lists **9** reserved identifiers: `conversation_id`, `session_id`, `thread_id`, `previous_response_id`, `cached_prompt_id`, `parent_message_id`, `run_id`, `context_id`, `assistant_id`.
- **CLAUDE.md** "Coding rules" section lists only **7** (missing `context_id` + `assistant_id`).
- **PR #19's AST rule** `willbuy/no-reserved-llm-identifiers` matches CLAUDE.md (7), so the lint passes through `context_id` / `assistant_id`.

Fix three things in one PR:
1. Update `packages/llm-adapter/eslint-rule.js` `FORBIDDEN` set to all 9 identifiers.
2. Add a test fixture for each new identifier (`context_id` + `assistant_id`).
3. Update `CLAUDE.md` "Coding rules" to list all 9 identifiers (fixing the seed of the drift).

Per CLAUDE.md "the spec is authoritative" — this is a real bug, just not blocking PR #19 merge.

Inline reviewer comment: https://github.com/NikolayS/willbuy/pull/19#issuecomment-4319645720

PR #179, opened 2026-04-26
feat(visitor-worker+shared): add tier_picked fields to prompt + VisitorOutput schema (#173)

### LINKED ISSUE #173: feat(visitor-worker): add tier_picked_if_buying_today + highest_tier_willing_to_consider to prompt output schema
## Problem

The visitor prompt (v0.1) does not ask the LLM to report which pricing tier the persona would pick. As a result:

1. report_json.tier_picked always shows 100% "none" — §5.18 #5 (tier-picked distribution chart) is useless.
2. scoreVisit() in packages/shared/src/scoring.ts has bump rules for bookmark_compare_later (+0.3) and start_free_hobby (+0.2) when the tier fields indicate a paid tier — these bumps never fire.
3. conv_score in the reports table is a plain mean of will_to_buy instead of the weighted scoreVisit() formula.

Evidence: apps/visitor-worker/src/prompt.ts output schema lists will_to_buy, next_action, and text fields only. packages/shared/src/scoring.ts defines PAID_TIERS and bump rules but they are never exercised because the fields are never populated.

## Fix

Add to the static prompt output schema:
- tier_picked_if_buying_today in {none, hobby, express, starter, scale, enterprise}
- highest_tier_willing_to_consider in {none, hobby, express, starter, scale, enterprise}

Update apps/aggregator/src/aggregator/main.py run_study() to pass these through scoreVisit() when computing conv_score.

## Acceptance criteria

- Visitor prompt includes both tier fields in the output spec
- At least one visit in the dogfood replay produces a non-none tier
- conv_score in reports uses scoreVisit(parsed, tier_today, considered) weighted formula
- tier_picked chart in the report shows a non-trivial distribution on a fresh run

## Notes

scoreVisit() is already correct — it just needs inputs. No changes to packages/shared/src/scoring.ts are required.

NikolayS:
Closes #173 (prompt side).

## Changes

- `packages/shared/src/visitor.ts` — adds `tier_picked_if_buying_today` and `highest_tier_willing_to_consider` to `VisitorOutput` schema, both optional with `default("none")` for backward compatibility with existing parsed rows
- `apps/visitor-worker/src/prompt.ts` — adds both fields to the static output spec so the LLM is asked to report them
- `apps/visitor-worker/test/helpers/fixtures.ts` — adds tier fields to `VALID_VISITOR_OUTPUT` fixture (required by schema default behavior)

## Why

The visitor prompt (v0.1) never asked the LLM for pricing tier intent. As a result:
- `report_json.tier_picked` showed 100% "none" in the dogfood run (useless distribution chart)
- `scoreVisit()` bump rules never fired
- `conv_score` used plain WTB mean instead of weighted formula

## What is NOT in this PR

The aggregator `scoreVisit()` wiring (reading tier fields to compute `conv_score`) is a follow-on. This PR lands the data collection side; the computation side follows once #168 merges.

## Test plan

- `bun run test` in `apps/visitor-worker` — 22/22 passing
- `bun run test` in `apps/web` — 106/106 passing
- Existing parsed rows without tier fields still validate (schema uses `.default("none")` + `.passthrough()`)

### DIFF
--- a/apps/api/src/routes/reports.ts
+++ b/apps/api/src/routes/reports.ts
@@ -167,6 +167,7 @@ interface ReportRow {
   scores_json: object | null;
   paired_tests_disagreement: boolean | null;
   ready_at: Date;
+  report_json: unknown | null;
 }
 
 // ---------------------------------------------------------------------------
@@ -191,7 +192,8 @@ export async function registerReportsRoutes(
         `SELECT r.id, r.study_id, r.share_token_hash, r.public,
                 r.expires_at, r.conv_score, r.paired_delta_json,
                 r.clusters_json, r.scores_json,
-                r.paired_tests_disagreement, r.ready_at
+                r.paired_tests_disagreement, r.ready_at,
+                r.report_json
            FROM reports r
            JOIN studies s ON s.id = r.study_id
           WHERE r.study_id = $1`,
@@ -219,6 +221,7 @@ export async function registerReportsRoutes(
           scores_json: report.scores_json,
           paired_tests_disagreement: report.paired_tests_disagreement,
           ready_at: report.ready_at.toISOString(),
+          report_json: report.report_json ?? null,
         });
       };
 
--- a/apps/visitor-worker/src/prompt.ts
+++ b/apps/visitor-worker/src/prompt.ts
@@ -43,6 +43,8 @@ const STATIC_PREFIX = [
   '  next_action ∈ {purchase_paid_today, contact_sales, book_demo,',
   '                 start_paid_trial, bookmark_compare_later,',
   '                 start_free_hobby, ask_teammate, leave}',
+  '  tier_picked_if_buying_today ∈ {none, hobby, express, starter, scale, enterprise}',
+  '  highest_tier_willing_to_consider ∈ {none, hobby, express, starter, scale, enterprise}',
   'Do NOT include commentary outside the JSON object. Do NOT wrap the',
   'JSON in code fences. Output exactly one valid JSON object.',
 ].join('\n');
--- a/apps/visitor-worker/test/helpers/fixtures.ts
+++ b/apps/visitor-worker/test/helpers/fixtures.ts
@@ -32,6 +32,8 @@ export const VALID_VISITOR_OUTPUT: VisitorOutputT = {
   confidence: 8,
   reasoning:
     'Strong pricing-page clarity but missing compliance signals; for a regulated fintech buyer the next step is sales.',
+  tier_picked_if_buying_today: 'starter',
+  highest_tier_willing_to_consider: 'scale',
 };
 
 export function validVisitorJsonString(): string {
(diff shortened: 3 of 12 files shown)


NikolayS:
## VERDICT: APPROVE

**Summary:** PR correctly adds `tier_picked_if_buying_today` and `highest_tier_willing_to_consider` to the VisitorOutput schema and prompt, with backward-compat defaults, and bundles three related fixes (report_json passthrough, single-variant null guards, schema length relaxation) that are all sound.

**Findings:**

- [PASS] **Schema backward compat** — both new fields use `.default('none')` on a `.passthrough()` schema, so existing stored rows without these keys still parse cleanly. Correct.

- [PASS] **Prompt wiring** — fields are added to `STATIC_PREFIX` in `prompt.ts` in the correct position (inside the output spec block, before the "Do NOT include commentary" line). The enum values match the Zod schema exactly.

- [PASS] **Fixture update** — `VALID_VISITOR_OUTPUT` in fixtures.ts gets concrete non-none values (`'starter'` / `'scale'`), which is a better test signal than leaving them as `'none'`.

- [PASS] **scoring.test.ts baseVisit** — `tier_picked_if_buying_today: 'none'` and `highest_tier_willing_to_consider: 'none'` added; test file compiles and all cases continue to cover the scoring logic.

- [PASS] **report_json passthrough (API)** — `report_json` added to the `ReportRow` interface, to the SELECT, and to the `send200()` response body. The `?? null` coercion is safe (JSONB column is nullable until the aggregator writes it).

- [PASS] **fetchReport.ts real-API path** — cookie forwarding uses the correct `wb_rt_${slug}` cookie name matching the §2 #20 spec. The `try/catch` around `cookies()` correctly handles the unit-test out-of-request-scope throw. Network errors are caught and return null. `body['report_json'] ?? null` is the right extraction key.

- [PASS] **PersonaGrid null guards** — `score_b !== null` guards on the "B:" score span and the entire verdict-B drawer block prevent "B: null" / "Variant B · score null" rendering on single-variant reports.

- [PASS] **Sort/swing null safety** — `(p.score_b ?? p.score_a) - p.score_a` means single-variant personas sort with swing=0, which groups them all at the bottom after any two-variant personas. Correct tie-breaking falls back to `backstory_id`.

- [NON-BLOCKING] **Swing badge still renders on single-variant cards** — when `score_b` is null, swing computes to 0 and the badge renders a gray "0". It does not crash, but it is visually noisy — the card implies a measurable delta when there is none. A `score_b !== null &&` guard around the badge span would clean this up, but it is cosmetic and safe to do in a follow-on.

- [PASS] **Schema length relaxation** (`histograms`, `next_actions`, `tier_picked`: `.length(2)` → `.min(1).max(2)`) — required for single-variant studies. The new test in `report-viz.test.tsx` exercises `Report.parse()` and `ReportView` render with a length-1 path; adequate coverage.

- [PASS] **Migration** — `0020_report_json.sql` uses `ADD COLUMN IF NOT EXISTS` (idempotent). Sqitch deploy wraps in a transaction and inserts the `_migrations` sentinel row. Plan entry is correctly chained on `0019_api_keys_label`.

**Manual test evidence:** present — PR description cites `bun run test` results: 22/22 in visitor-worker, 106/106 in web.

PR #181, opened 2026-04-26
fix(aggregator): small-dataset fallback when HDBSCAN finds no clusters (#180)

### LINKED ISSUE #180: fix(aggregator): HDBSCAN min_cluster_size=3 produces empty clusters on small dogfood datasets
## Problem

In the Sprint 6 dogfood run (study_id=1, n=10 visits), `unanswered_blockers` produced **zero clusters** despite each visit contributing 2–4 blocker strings (~30 total). The other finding kinds (questions, confusions, objections) also showed thin clusters or empty results in practice.

Root cause: `cluster.py` uses `HDBSCAN(min_cluster_size=3, min_samples=3)`. With ~30 semantically diverse strings from only 10 visits, the density required to form a cluster of size ≥3 is rarely met. Strings that would be meaningfully grouped (e.g., multiple variants of "no transparent pricing for enterprise") fall just outside the cosine-distance threshold and get labeled as noise.

The early bail-out at `len(normalized) < 3` compounds this for finding kinds with fewer than 3 unique strings.

## Impact

Sprint 7 §5.18 report display will show empty cluster sections for most finding kinds on studies with < ~50 visits. The dogfood report_json has `unanswered_blockers: []` even though the raw per-visit data is rich.

## Proposed fix

Add a **small-dataset fallback** inside `cluster_findings()`: if HDBSCAN returns no clusters but `normalized` is non-empty, return a single synthetic cluster containing all strings (id=0). This preserves the density-based clustering for large datasets and degrades gracefully for small ones.

Threshold suggestion: apply fallback when `len(normalized) < 20` OR when HDBSCAN returns no clusters and `len(normalized) > 0`.

No spec amendment needed — spec §17 only mandates the HDBSCAN parameters when a real cluster is formed; the fallback path is additive.

## Acceptance

- `cluster_findings(['a', 'b', 'c', 'd'])` returns one cluster with all 4 members (currently returns `[]`)  
- Existing clustering tests still pass (large-N cases unchanged)
- Dogfood aggregator re-run on study_id=1 produces non-empty `unanswered_blockers` clusters

NikolayS:
## Problem

`cluster_findings()` uses `HDBSCAN(min_cluster_size=3)`. With sparse finding strings from a small study (< ~50 visits), the density-based pass labels everything as noise and returns an empty list. In the Sprint 6 dogfood (study_id=1, n=10 visits), `unanswered_blockers` produced zero clusters despite ~30 total strings.

## Fix

When HDBSCAN returns no clusters but `len(normalized) >= 3`, return one catch-all cluster containing all normalized strings. Large datasets are unaffected (they get real density-based clusters). The early bail for `len < 3` is unchanged.

## Tests

- Existing test `cluster_findings(["alpha", "beta"]) == []` still passes (early bail, < min_cluster_size)
- New `test_cluster_findings_small_dataset_fallback`: monkeypatches `_embed` with random-spread vectors that HDBSCAN cannot cluster, asserts one catch-all cluster is returned

## Closes

Fixes #180

### DIFF
--- a/apps/aggregator/src/aggregator/cluster.py
+++ b/apps/aggregator/src/aggregator/cluster.py
@@ -164,6 +164,21 @@ def cluster_findings(strings: Sequence[str]) -> list[Cluster]:
         key=lambda g: g[0],  # tie-break by lowest sorted-input index
     )
 
+    # Small-dataset fallback (issue #180): when HDBSCAN finds no clusters but
+    # strings exist, return one catch-all cluster rather than an empty list.
+    # Prevents hollow report sections for finding kinds with sparse data (< ~50
+    # visits); HDBSCAN needs density ≥ min_cluster_size to form any cluster, so
+    # small studies always fall through to noise. The fallback fires only when
+    # the density-based pass found nothing — large datasets are unaffected.
+    if not groups:
+        return [
+            Cluster(
+                id=0,
+                members=tuple(normalized),
+                member_indices=tuple(range(len(normalized))),
+            )
+        ]
+
     return [
         Cluster(
             id=new_id,
--- a/apps/aggregator/tests/test_cluster.py
+++ b/apps/aggregator/tests/test_cluster.py
@@ -81,8 +81,41 @@ def test_cluster_findings_returns_lex_sorted_inputs_in_member_indices() -> None:
 
 def test_cluster_findings_handles_empty_and_tiny_inputs() -> None:
     assert cluster_findings([]) == []
-    # 2 inputs cannot meet min_cluster_size=3 → all noise → 0 clusters.
+    # 2 inputs: early bail (< min_cluster_size=3) before HDBSCAN → no fallback.
     assert cluster_findings(["alpha", "beta"]) == []
+
+
+def test_cluster_findings_small_dataset_fallback(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Issue #180: when HDBSCAN returns no clusters, return one catch-all cluster.
+
+    With sparse strings in a small study HDBSCAN with min_cluster_size=3 labels
+    everything as noise. The fallback must surface a single cluster with all
+    normalized strings rather than an empty list.
+    """
+    import numpy as np
+    import aggregator.cluster as cluster_mod
+
+    strings = [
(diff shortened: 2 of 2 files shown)


NikolayS:
## VERDICT: APPROVE

**Summary:** The small-dataset fallback is correctly placed, correctly scoped, and the new test exercises it adequately; no runtime bugs or logic errors found.

**Findings:**

- [PASS] **Fallback placement and variable scope** — `groups` is built from `by_label` (noise label `-1` already excluded), so `not groups` fires exactly when HDBSCAN labeled everything noise. `normalized` is still in scope and holds the post-dedup lex-sorted strings. Correct.

- [PASS] **Implicit `len >= 3` guard** — the early bail at line 111 (`len(normalized) < 3 → return []`) guarantees `len(normalized) >= 3` is already true by the time the fallback is reached. No secondary length check needed in the fallback.

- [PASS] **`member_indices=tuple(range(len(normalized)))`** — `normalized` is lex-sorted by construction, so indices 0..N-1 are in ascending order. This satisfies the invariant checked by `test_cluster_findings_returns_lex_sorted_inputs_in_member_indices` (each group's members must be in sorted index order).

- [PASS] **`id=0` and `members=tuple(normalized)`** — consistent with the regular path's `enumerate(groups)` starting at 0. All strings are included (there are no noise points to exclude when no density cluster formed).

- [PASS] **Determinism** — fallback path calls no randomized code; output is fully determined by the normalized (sorted, deduped) input.

- [PASS] **Existing tests unaffected** — `cluster_findings([]) == []` (len 0, early bail) and `cluster_findings(["alpha", "beta"]) == []` (len 2, early bail) both still return `[]` because the fallback is never reached. Comment in the test was updated accurately.

- [NON-BLOCKING] **No test for `len(normalized) == 3` hitting the fallback** — the new test uses 4 strings; a 3-string variant (the minimum that escapes the early bail) would complete boundary coverage. The 4-string test already exercises the code path end-to-end, so this is a nice-to-have, not blocking.

**Manual test evidence:** absent (PR description references the Sprint 6 dogfood study as the motivation but does not include a before/after screenshot or log excerpt from study_id=1).

PR #187, opened 2026-04-26
fix(aggregator): set meta.slug = study_id, not share_token_hash (#184)

### LINKED ISSUE #184: fix(aggregator+spec): meta.slug in report_json should be study_id, not share_token_hash
## Problem (audit D2)

Sprint 7 spec alignment audit found a naming confusion: `report_json.meta.slug` currently stores the 64-char SHA-256 `share_token_hash` hex digest. The field name `slug` implies it is the URL identifier for `/r/<slug>`, but the actual route resolves `<slug>` as `study_id`.

This means the `meta.slug` shown in the debug panel (`Study <code>{report.meta.slug}</code>`) displays a 64-char hex string instead of a meaningful study identifier.

## Fix

Change `_build_report_json()` to set:
```python
meta = {
    "slug": str(study_id),   # the /r/:slug URL identifier
    "low_power": len(visits) < 20,
}
```

`share_token_hash` is stored as a separate column on the `reports` row and does not need to appear in the visualization JSON.

## Acceptance

- `report_json.meta.slug` equals the study_id string (e.g. `"1"` for dogfood study)
- `/r/1` debug panel shows `Study 1`, not the hex hash
- Existing e2e test updated to assert `meta.slug == str(study_id)`

NikolayS:
## Summary

- `_build_report_json()` was setting `meta.slug = share_token_hash` (a 64-char hex string), but the `/r/<slug>` URL route uses `study_id` as the key
- Added `study_id: str` parameter to `_build_report_json()` and changed the slug value to `str(study_id)`
- Updated `run_study()` to pass `study_id` through to `_build_report_json()`
- Updated the e2e test assertion in `test_main_e2e.py` to verify `meta.slug == "study_e2e_001"` (the study_id) instead of the hash

## Test plan

- [x] `cd apps/aggregator && .venv/bin/python -m pytest tests/ -x -q` — all tests pass except the pre-existing `test_paired_delta_disagreement_true` failure (expected per issue)

Fixes #184

### DIFF
--- a/apps/aggregator/src/aggregator/main.py
+++ b/apps/aggregator/src/aggregator/main.py
@@ -237,10 +237,11 @@ def _build_report_json(
     clusters: dict[str, list[dict]],
     backstory_map: dict[int, dict],
     share_token_hash: str,
+    study_id: str,
 ) -> dict:
     """Compute the full §5.18 Report visualization blob."""
     meta = {
-        "slug": share_token_hash,
+        "slug": str(study_id),
         "low_power": len(visits) < 20,
     }
 
@@ -439,6 +440,7 @@ def run_study(
         clusters=clusters,
         backstory_map=backstory_map,
         share_token_hash=share_token_hash,
+        study_id=study_id,
     )
 
     db.execute(
--- a/apps/aggregator/tests/test_main_e2e.py
+++ b/apps/aggregator/tests/test_main_e2e.py
@@ -157,8 +157,8 @@ def test_e2e_run_study_writes_report_and_clusters(tmp_path: Path) -> None:
     n_label_rows = cur.fetchone()[0]
     assert n_label_rows == total_clusters
 
-    # report_json slug matches share_token_hash.
-    assert rj["meta"]["slug"] == report_row[2]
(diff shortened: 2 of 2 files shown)


NikolayS:
## VERDICT: APPROVE

**Summary:** Fixes `meta.slug` in the report JSON to use `study_id` (the actual URL route key) instead of `share_token_hash`. Change is surgically small — one new parameter on `_build_report_json()`, one assignment change, one pass-through in `run_study()`, and one updated test assertion.

**Findings:**
- [PASS] `_build_report_json()` now accepts `study_id: str` as an explicit parameter — clean API, no reliance on closure state.
- [PASS] `meta["slug"]` is set to `str(study_id)`, which is safe even if `study_id` is already a string (idempotent cast).
- [PASS] `run_study()` correctly passes `study_id=study_id` through to the updated signature.
- [PASS] Test assertion updated from checking the hash (`report_row[2]`) to checking the literal `"study_e2e_001"` — directly verifies the correct value is written, not just that it matches whatever was passed.
- [PASS] No other callers of `_build_report_json()` in the diff; the function is internal so no public API breakage.

PR #191, opened 2026-04-26
feat(api): wire Resend cap-warning email for §5.6 daily-cap threshold

### LINKED ISSUE #174: feat(web): wire Resend transactional email for §5.6 daily-cap warning
apps/api/src/billing/cap-warning.ts has a TODO stub at line ~61: the 50% daily-cap warning email is recorded in the DB but never actually sent.

## What is needed

Wire Resend (or equivalent transactional provider) to send the cap-warning email when sendCapWarningIfDue() returns true. The spec §5.6 says: "operator receives an email when daily spend crosses 50% of DAILY_CAP_CENTS".

## Acceptance criteria

- cap-warning email is sent via Resend when the 50% threshold is crossed
- RESEND_API_KEY env var wired in api service config
- Email includes: account ID, current spend, cap value, study ID that triggered it
- cap_warning_sent_at is still set before the send so re-entry is idempotent even if send fails
- Integration test or stub-verified unit test covers the send path

NikolayS:
## Summary

- Adds `CapWarningEmailOptions` interface and `sendCapWarning` method to `ResendClient` in `resend.ts`; implements stub (console.log) and live (Resend API with HTML table body) variants
- Extends `MaybeWarnCapInput` with `owner_email`, `study_id`, `resend`; `maybeWarnCap` now calls `sendCapWarning` after the exactly-once `cap_warnings` INSERT — errors are caught/logged, not rethrown, preserving the study creation response
- Threads a `postgres` sql client + `resend` through `buildServer` → `registerStudiesRoutes`; `POST /studies` calls `maybeWarnCap` (fire-and-forget via `void`) after COMMIT

## Tests

- Updated existing AC5 test to pass new required fields; all 6 pre-existing tests still pass
- Added AC7: `sendCapWarning` called with correct params when 50% threshold crossed
- Added AC8: `sendCapWarning` not called when below threshold
- Added AC9: `sendCapWarning` not called when idempotency row already exists (ON CONFLICT)
- Updated all `buildStubResend` stubs in `auth.test.ts`, `dashboard.test.ts`, `api-keys.test.ts`, `studies-list.test.ts` to implement the new interface method

## Test plan

- [x] `bun run typecheck` — zero errors
- [x] `bun run test test/atomic-spend.test.ts` — 9/9 passed (Docker required, confirmed green locally)
- [ ] CI green on full suite

Closes #174

### DIFF
--- a/apps/api/src/billing/cap-warning.ts
+++ b/apps/api/src/billing/cap-warning.ts
@@ -8,13 +8,12 @@
  * inserts with a unique_violation, so only one concurrent caller succeeds.
  *
  * Returns true when this caller was the one that inserted the warning row
- * (i.e., should enqueue the email). Returns false when the row already
+ * (i.e., the email was dispatched). Returns false when the row already
  * existed (another caller beat us, or the 50% threshold hasn't been crossed).
- *
- * Email sending is stubbed — TODO wire Resend when it's available.
  */
 
 import type postgres from 'postgres';
+import type { ResendClient } from '../email/resend.js';
 
 export type MaybeWarnCapInput = {
   sql: ReturnType<typeof postgres>;
@@ -24,6 +23,12 @@ export type MaybeWarnCapInput = {
   /** The new total cents after a successful reserve (from RETURNING cents). */
   new_cents: number;
   daily_cap_cents: number;
+  /** Account owner email — recipient of the cap-warning email. */
+  owner_email: string;
+  /** The study that pushed spend over the threshold. */
+  study_id: string;
+  /** Resend client for sending the warning email. */
+  resend: ResendClient;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


NikolayS:
## VERDICT: APPROVE

**Summary:** Wires the Resend cap-warning email into `maybeWarnCap` with correct idempotency and error-handling semantics, and fills out the test matrix to AC5/7/8/9.

**Findings:**
- [PASS]: idempotency preserved — `sendCapWarning` is called only after the `INSERT … ON CONFLICT DO NOTHING` returns a row (i.e., `insertedRows.length === 1`); if the row already existed the function returns `false` before reaching the send path
- [PASS]: email failures are caught in a try/catch and logged via `console.error`; nothing is rethrown, so the study-creation response is never broken by a transient Resend error
- [PASS]: `sendCapWarning` no-op stub added to all four test files that build a `ResendClient` (api-keys, auth, dashboard, studies-list); atomic-spend test gets a recording stub (`CapWarnStub`) for assertion purposes
- [PASS]: AC7 (threshold crossed → call with correct params), AC8 (below threshold → no call), AC9 (idempotent repeat → still exactly one call), and the existing AC5 concurrency test are all covered

PR #228, opened 2026-04-27
fix(web): link pending report page to study status page

NikolayS:
## Summary

- When a user visits `/r/<slug>` while the study is still aggregating, they previously saw a dead-end "Refresh in a moment" message with no path forward
- The reports API already returns `study_id` in the 200 response even when `report_json` is null (pending state)
- Now `fetchReport` surfaces `studyId` in the `PendingReport` discriminant, and the page renders a "Track progress on the study status page →" link pointing to `/dashboard/studies/:id`
- Falls back to "Refresh in a moment" text if `study_id` is absent

## Test plan

- [x] `pages.test.tsx` — added pending-state test with stubbed fetch; asserts the study link is present
- [x] Full web test suite: 111 passed, 0 failed


NikolayS:

PR #229, opened 2026-04-27
fix(web): encode pack param inside redirect path on pricing CTA

NikolayS:
## Summary

- Unauthenticated "Buy →" on `/pricing` was generating:
  `/sign-in?redirect=/pricing&pack=starter`
- After sign-in, the verify handler reads only `redirect=` — the `pack=starter` was a separate sibling param and got discarded
- User landed on `/pricing` without the pack param (benign but inconsistent)

**Fix**: encode the full redirect path so `pack=` is preserved:
  `/sign-in?redirect=%2Fpricing%3Fpack%3Dstarter`

The `safeRedirect()` helper in `auth.ts` allows query strings in redirect paths (only blocks non-relative paths and `//`), so this works end-to-end.

## Test plan

- [x] Added pricing test asserting encoded URL form for all 3 packs
- [x] Full web test suite: 112 passed, 0 failed

### DIFF
--- a/apps/web/app/pricing/PricingCta.tsx
+++ b/apps/web/app/pricing/PricingCta.tsx
@@ -16,9 +16,10 @@ export function PricingCta({ packId, label, usd, isAuthenticated }: PricingCtaPr
   if (isAuthenticated) {
     return <BuyButton packId={packId} label={label} usd={usd} />;
   }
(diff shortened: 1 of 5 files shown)


NikolayS:

PR #231, opened 2026-04-27
test(api): assert report_public field in GET /studies/:id responses

NikolayS:
## Summary

PR #227 added `report_public` to `GET /studies/:id` (and the session-cookie mirror) via a LEFT JOIN with reports, but shipped no tests for the new field.

Adds 3 acceptance tests:
- report exists and `public = false` → response includes `report_public: false`
- report exists and `public = true` → response includes `report_public: true`
- no report row yet (study still in progress) → `report_public` absent from response body

## Test plan

- [x] All 3 new tests pass against real DB
- [x] Full API test suite: 148 passed, 0 failed


NikolayS:

PR #240, opened 2026-04-27
test(web): pending-report page state coverage

NikolayS:
## Summary

`GET /r/:slug` had no test for the pending state (API returns `report_json: null`). The only coverage was the 'not found' path and the happy-path fixture demo slug.

Adds one test that stubs `fetch()` to return `{ report_json: null, urls: null }` (same shape as the API when the aggregator hasn't run yet) and asserts the page renders the "Report is being prepared" message.

## Test plan

- [x] New test passes (`bun run test --run test/pages.test.tsx`)
- [x] All 6 pages tests pass


NikolayS:

PR #21, opened 2026-04-25
feat(visitor-worker): orchestrate backstory + page + LLMProvider with retries

NikolayS:
Closes #9

## Summary

`apps/visitor-worker` ships `runVisit(opts) -> VisitResult`: builds the visitor prompt, calls `LLMProvider.chat`, validates against `VisitorOutput` zod from `@willbuy/shared`, and on validation failure issues a fresh-context schema-repair retry up to twice — each retry a NEW logical-request key, prior bad output passed back as user-role content only (NEVER as an assistant-role turn). On adapter `status: 'error'` (transport failure) the orchestrator returns `failed/transport` immediately — transport retries are the adapter's responsibility per spec §5.15.

## Spec link

Implements **spec §2 #14** (up to 2 schema-repair retries; fresh-context — no carried-over assistant turns) + **spec §2 #15** (strict `VisitorOutput` schema, `max_tokens=800`) + **spec §5.15** (logical-request key vs transport attempts; `repair_generation` increments → new logical key; transport retries reuse the key inside the adapter; orchestrator does not transport-retry).

Inherits **amendment A1** automatically through the `VisitorOutput` zod from `@willbuy/shared` — no `next_action` enum literal appears in this package.

## Files

- `apps/visitor-worker/src/index.ts` — package barrel.
- `apps/visitor-worker/src/visitor.ts` — `runVisit` orchestrator + `computeLogicalRequestKey`.
- `apps/visitor-worker/src/prompt.ts` — `buildStaticPrefix` (deterministic, byte-identical across visits → cacheable prefix invariant), `buildDynamicTail`, `buildRepairTail` (single user-role payload with explicit `PRIOR_BAD_OUTPUT_BEGIN` / `PRIOR_BAD_OUTPUT_END` markers; defense-in-depth grep asserts no "assistant" string in the repair tail).
- `apps/visitor-worker/test/` — 6 test files, 10 tests; `MockProvider` test double records every chat call and replays scripted responses.

## Test evidence — red → green commit history

Linear, no merges:

```
e007160 visitor-worker: refactor — drop stale eslint-disable in test mock
49b4b56 visitor-worker: green — acc#5 promote computeLogicalRequestKey to package barrel
1cb9e67 visitor-worker: red — acc#5 publicly export computeLogicalRequestKey + key stability
22fb865 visitor-worker: green — acc#4 transport-error early return, no schema repair
3027c25 visitor-worker: red — acc#4 chat() error → failed/transport, attempts=1, no schema repair
[...]

PR #241, opened 2026-04-27
test(api): AC9 — 2 active API key cap returns 409

NikolayS:
## Summary

The DB trigger enforcing the ≤2 active-key cap per account (spec §2 #21) had no integration test — the rule was documented in the spec but the API's 409 response path was never exercised.

Adds AC9: creates 2 keys for a fresh isolated account (both return 201), then attempts a third key and asserts:
- Response is 409
- Error message matches "maximum of 2 active"

## Test plan

- [x] AC9 passes (`bun run test --run test/api-keys.test.ts`)
- [x] All 9 api-keys tests pass (was 8)


NikolayS:

PR #242, opened 2026-04-27
test(web): pricing page authenticated-state coverage

NikolayS:
## Summary

The pricing page is auth-aware: signed-in users see `BuyButton` (direct Stripe checkout) instead of the sign-in redirect CTA. Only the unauthenticated state was tested.

Adds one test that overrides the `next/headers` mock to simulate a session cookie present (`has: () => true`) and asserts:
- `BuyButton` renders ("Buy →" text visible)
- No `/sign-in?redirect=/pricing` href is emitted (the unauthenticated CTA)

## Test plan

- [x] New test passes (`bun run test --run test/pricing.test.tsx`)
- [x] All 11 pricing tests pass (was 10)

### DIFF
--- a/apps/web/test/pricing.test.tsx
+++ b/apps/web/test/pricing.test.tsx
@@ -74,4 +74,17 @@ describe('/pricing page (issue #144)', () => {
     const html = await getHtml();
     expect(html).toMatch(/sign-in/i);
   });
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #245, opened 2026-04-27
test(api): SESSION_HMAC_KEY + SHARE_TOKEN_HMAC_KEY env validation

NikolayS:
## Summary

Both `SESSION_HMAC_KEY` and `SHARE_TOKEN_HMAC_KEY` have a 32-character minimum enforced by Zod at startup, but neither was exercised in `env.test.ts`. A misconfiguration (e.g., short key in `.env.local`) would fail at runtime with a cryptic Zod error rather than being caught at test time.

Adds 3 tests:
- Short `SESSION_HMAC_KEY` → throws with message matching `/SESSION_HMAC_KEY/`
- Short `SHARE_TOKEN_HMAC_KEY` → throws with message matching `/SHARE_TOKEN_HMAC_KEY/`
- Valid 32-char values for both → accepted and passed through

## Test plan

- [x] 3 new tests pass (`bun run test --run test/env.test.ts`)
- [x] All 10 env tests pass (was 7)


NikolayS:

Issue #74, opened 2026-04-25
chore(web): rev-pr72 nits — clear poll on terminal status + fix /r/:slug URL


NikolayS:
Two non-blocking findings from rev-pr72 review of merged PR #72:

**MINOR-1: Polling continues after terminal status.**
`apps/web/app/dashboard/studies/[id]/page.tsx`'s `setInterval(pollStatus, 5000)` is only cleared on component unmount, not when `status === 'ready'` or `status === 'failed'`. After a study finishes, the page keeps polling `GET /studies/:id` every 5s for as long as the user keeps the tab open. Easy fix: in the polling effect, check `if (status === 'ready' || status === 'failed') return` before the next interval, OR `clearInterval` inside the handler when terminal is reached.

**MINOR-2: `/r/:slug` fallback URL is wrong.**
The status page falls back to `/r/study-${s.id}` when no slug is returned by `GET /studies/:id`, but the report route from PR #63 expects a bare slug — likely the study's actual slug field, OR (placeholder for Sprint 3) just `${s.id}`. The current fallback would 404. Fix: either expose `slug` from the `GET /studies/:id` response (server-side change in `apps/api/src/routes/studies.ts`), OR make the fallback `/r/${s.id}` to match what `apps/web/app/r/[slug]/page.tsx` resolves.

Source: rev-pr72 review at https://github.com/NikolayS/willbuy/pull/72#issuecomment-4320377139

PR #246, opened 2026-04-27
test(web): polling stops at terminal status (issue #74 MINOR-1)

NikolayS:
## Summary

The fix from issue #74 MINOR-1 that clears the `setInterval` when study status reaches `'ready'` or `'failed'` had no regression test. Without it, the page would keep hitting `GET /api/studies/:id` every 5 s indefinitely while the tab is open.

Adds one test:
- First poll response returns `status: 'ready'` (terminal)
- Fake timers are advanced 3× the 5 s poll interval (15 s)
- Asserts `fetch` call count did NOT increase after the initial poll
- Asserts UI still shows "Ready", not "Capturing" from the sentinel second-response

## Test plan

- [x] New test passes (`bun run test --run test/studies-new.test.ts`)
- [x] All 9 studies-new tests pass (was 8)

### DIFF
--- a/apps/web/test/studies-new.test.ts
+++ b/apps/web/test/studies-new.test.ts
@@ -338,4 +338,72 @@ describe('StudyStatusPage — polling', () => {
     });
     expect(fetchSpy.mock.calls.length).toBeGreaterThan(callCount);
   }, 15_000);
+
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #250, opened 2026-04-27
test(shared): Report Zod schema validation

NikolayS:
## Summary

The `Report` Zod schema (`packages/shared/src/report.ts`) is the contract between the aggregator and the report page, but had no dedicated unit tests — it was only exercised as a side effect of `report-viz.test.tsx`. A schema regression (e.g., wrong constraint on a field) could silently break the aggregator-web wire without being caught.

Adds 10 tests in `packages/shared/test/report.test.ts`:
- Canonical web fixture parses without errors
- Missing `meta.slug` → rejected
- Invalid `headline.verdict` enum → rejected
- `headline.paired_t_p` out of [0,1] → rejected
- `histograms` with 0 entries (min 1) → rejected
- `histograms` with 3 entries (max 2) → rejected
- Histogram `bins` array with 10 entries (needs 11) → rejected
- `paired_dot.swing` not in enum → rejected
- `score_b: null` / `verdict_b: null` on personas → accepted (single-variant)
- `persona.score_a > 10` → rejected

## Test plan

- [x] 10 new tests pass (`bun run test --run test/report.test.ts` from `packages/shared/`)

### DIFF
--- a/packages/shared/test/report.test.ts
+++ b/packages/shared/test/report.test.ts
@@ -1,0 +1,115 @@
+/**
+ * report.test.ts — Zod schema validation for the §5.18 Report wire shape.
+ *
+ * The Report schema is the contract between the aggregator and the report
+ * page. These tests assert that:
+ *   1. The web fixture (used as the canonical happy-path) parses correctly.
+ *   2. Key constraint violations are caught (required fields, min/max,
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #251, opened 2026-04-27
test(aggregator): unit tests for helper functions

NikolayS:
## Summary

Three aggregator helper functions are exercised indirectly by `test_main_e2e.py` but had no isolated unit tests. Adding them makes the invariants explicit and provides a faster feedback loop for regressions.

### Functions covered

**`_build_visits_by_backstory`** — pairs A+B visit rows by backstory ID, drops incomplete pairs:
- Paired A+B → included
- Only A (no B partner) → dropped
- Empty input → `{}`
- All unpaired → `{}`

**`_collect_findings`** — aggregates finding strings from all visits, filtering non-strings:
- Aggregates strings from multiple visits
- Filters out `None`, integers (defensive against LLM returning mixed types)
- Missing kind key in output → category stays empty
- `null` list for kind (LLM quirk) → treated as empty
- Empty input → all categories empty

**`_bs_map_get`** — backstory map lookup tolerating non-integer IDs:
- Integer key → found
- String `"1"` → coerced to int, found
- Missing key → `{}`
- Non-numeric string/`None` → `{}` (no raise)

## Test plan

- [x] 13 new tests pass in Docker CI (same environment as existing aggregator tests)

### DIFF
--- a/apps/aggregator/tests/test_helper_functions.py
+++ b/apps/aggregator/tests/test_helper_functions.py
@@ -1,0 +1,128 @@
+"""Unit tests for aggregator helper functions.
+
+These functions are exercised indirectly by test_main_e2e.py but benefit
+from isolated unit tests that make the invariants explicit:
+  - _build_visits_by_backstory: drops incomplete pairs (no single-variant visits)
+  - _collect_findings: filters non-string entries, handles missing kinds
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #253, opened 2026-04-27
test(web): api-client direct unit tests

NikolayS:
## Summary

`getStudy()` and `createStudy()` in `lib/api-client.ts` were only tested indirectly via component tests. The schema-mismatch code path (`Response shape mismatch: ...`) and the HTTP-status-text fallback when there's no JSON `error` field were entirely untested.

Adds 7 direct tests covering both functions:
- **`getStudy`**: valid 200 → `{ ok: true }`, network error → `{ status: 0 }`, 401 with JSON error body, schema mismatch on 200 response → `{ ok: false, error: 'Response shape mismatch: ...' }`
- **`createStudy`**: 201 created, 402 cap exceeded (error from body), 500 without JSON body (falls back to `HTTP 500`)

## Test plan

- [x] 7 new tests pass (`bun run test --run test/api-client.test.ts`)

### DIFF
--- a/apps/web/test/api-client.test.ts
+++ b/apps/web/test/api-client.test.ts
@@ -1,0 +1,139 @@
+// @vitest-environment jsdom
+//
+// Direct unit tests for getStudy() and createStudy() in lib/api-client.ts.
+//
+// The functions are exercised indirectly by components, but the schema-mismatch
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #256, opened 2026-04-27
test(api): session HMAC encode/decode unit tests

NikolayS:
## Summary

`encodeSession()`, `decodeSession()`, and `cookieName()` from `auth/session.ts` are used as test helpers in nearly every integration test, but were never isolated. The critical tamper-resistance and expiry invariants had no dedicated assertion.

Adds 10 unit tests:
- `cookieName()` returns `__Host-wb_session` in production and `wb_session` in test/dev
- Valid payload round-trips correctly
- Expired session → `null`
- Tampered MAC → `null`
- Wrong key → `null`
- No dot separator → `null`
- Empty string → `null`
- Invalid JSON payload (malformed base64) → `null`
- Distinct keys produce distinct tokens

## Test plan

- [x] 10 new tests pass (`bun run test --run test/session.test.ts`)

### DIFF
--- a/apps/api/test/session.test.ts
+++ b/apps/api/test/session.test.ts
@@ -1,0 +1,112 @@
+/**
+ * session.test.ts — unit tests for encodeSession / decodeSession / cookieName
+ * (apps/api/src/auth/session.ts).
+ *
+ * These functions are used in nearly every integration test via test helpers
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #258, opened 2026-04-27
test(api): buildResendClient() stub mode unit tests

NikolayS:
## Summary

`buildResendClient()` (`email/resend.ts`) had no unit tests — it was only exercised indirectly via integration tests that inject custom stub implementations. The key invariants — placeholder key activates stub mode, `callCount` increments, `testMode=true` prevents real API calls — were never explicitly tested.

Adds 8 unit tests:
- No throw on `testMode=true`, placeholder key `re_not_configured`, or empty string
- `sendMagicLink` and `sendCapWarning` resolve in stub mode
- `callCount` starts at 0, increments after each `sendMagicLink`
- Placeholder key `re_not_configured` activates stub mode (no real Resend call → no throw)

## Test plan

- [x] 8 new tests pass (`bun run test --run test/resend.test.ts`)

### DIFF
--- a/apps/api/test/resend.test.ts
+++ b/apps/api/test/resend.test.ts
@@ -1,0 +1,76 @@
+/**
+ * resend.test.ts — unit tests for buildResendClient() (email/resend.ts).
+ *
+ * The module has two key invariants not covered by integration tests:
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #259, opened 2026-04-27
test(api): constantTimeEquals() unit tests

NikolayS:
## Summary

`constantTimeEquals()` is the timing-safe string comparator used to gate the `/metrics` endpoint (compares bearer tokens without short-circuiting on mismatch). It's security-critical but had no direct unit tests — only reachable indirectly through the Fastify integration suite.

Adds 9 unit tests:
- Equal non-empty strings → `true`
- Unequal strings (same length) → `false`
- Different lengths → `false` (both orderings)
- Two empty strings → `true`
- One empty, one non-empty → `false`
- Differ at last character only → `false`
- Differ at first character only → `false`
- Long identical strings (256 chars) → `true`
- Long strings differing at one byte → `false`

## Test plan

- [x] 9 new tests pass (`bun run test --run test/metrics-utils.test.ts`)

### DIFF
--- a/apps/api/test/metrics-utils.test.ts
+++ b/apps/api/test/metrics-utils.test.ts
@@ -1,0 +1,62 @@
+/**
+ * metrics-utils.test.ts — unit tests for constantTimeEquals() from metrics/registry.ts.
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #269, opened 2026-04-27
fix(api)+test: remove non-existent r.slug from /api/studies/:id; session-auth integration tests

NikolayS:
## Summary

- **Bug fix**: PR #227 introduced `r.slug` in the `GET /api/studies/:id` session-cookie route's SQL query, but the `reports` table has no `slug` column. Every call to this endpoint returned HTTP 500. Removed `r.slug` and aligned with the API-key sibling route which correctly uses `r."public"`.
- **Tests**: Adds a `describeIfDocker` integration test suite (issue #209) covering: 200 with study data for the owner, 401 without session cookie, 404 for cross-account isolation (no existence leak per spec §2 #20).

## Test plan

- [ ] CI passes (the 500 regression is confirmed fixed by the new integration tests)
- [ ] `GET /api/studies/:id` returns 200 in the dashboard when viewing a study

### DIFF
--- a/apps/api/src/routes/studies.ts
+++ b/apps/api/src/routes/studies.ts
@@ -366,11 +366,10 @@ export async function registerStudiesRoutes(
         status: string;
         created_at: Date;
         finalized_at: Date | null;
-        slug: string | null;
         report_public: boolean | null;
       }>(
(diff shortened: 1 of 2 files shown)


NikolayS:

PR #276, opened 2026-04-27
test(log): maskApiKey + maskEmail direct unit tests — 17 cases (spec §5.12)

NikolayS:
## Summary

17 direct unit tests for `maskApiKey` and `maskEmail` from `@willbuy/log`. These helpers are exported but were previously only tested indirectly via `redact()` integration assertions — no `describe` block of their own.

**`maskApiKey`** (6 tests): `***` prefix + last-4 format; 4-char, short, empty, long key, exact format.

**`maskEmail`** (11 tests): typical email, single-char local, multi-component TLD (`.co.uk`), domain-without-dot, no-`@`, `@` at position 0, `@` at last position, numeric local-part, long local-part, output does not leak original email.

## Test plan

- [ ] `bun run test` in `packages/log` — all 17 tests pass

### DIFF
--- a/packages/log/test/masks.test.ts
+++ b/packages/log/test/masks.test.ts
@@ -1,0 +1,96 @@
+/**
+ * masks.test.ts — direct unit tests for maskApiKey and maskEmail (spec §5.12).
+ *
+ * These pure helpers are exported from @willbuy/log but currently only tested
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #278, opened 2026-04-27
test(log): resolveSalt + shouldWriteToFile branch coverage — production-throw safety invariant

NikolayS:
## Summary

12 tests for the env-driven branch logic in `packages/log/src/index.ts` (`resolveSalt` + `shouldWriteToFile`). These private functions were **completely uncovered** — all existing log tests pass explicit `urlHashSalt` and `destination`, bypassing both functions entirely.

**`resolveSalt` (8 tests)**:
- Explicit non-empty `urlHashSalt` wins (no env var needed)
- `WILLBUY_LOG_HASH_SALT` env var wins when explicit is absent
- Empty env var falls through to dev fallback in non-production
- `NODE_ENV=development`, no explicit salt, no env var → dev fallback, no throw
- **`NODE_ENV=production` + no explicit + no env var → throws `'WILLBUY_LOG_HASH_SALT must be set in production'`** (safety invariant — was completely untested)
- Production with explicit salt → no throw
- Production with env var → no throw

**`shouldWriteToFile` (4 tests)**:
- Explicit `destination` overrides file-vs-stdout regardless of `WILLBUY_LOG_TO_FILE`
- `WILLBUY_LOG_TO_FILE=0` → non-file path
- `NODE_ENV=production` without the flag → `shouldWriteToFile` returns true (verified via the salt-missing throw that fires before the file-open)

Each test snapshots and restores `process.env` to avoid cross-test pollution.

## Test plan

- [ ] `bun run test` in `packages/log` — all 12 tests pass

### DIFF
--- a/packages/log/test/build-logger-env.test.ts
+++ b/packages/log/test/build-logger-env.test.ts
@@ -1,0 +1,204 @@
+/**
+ * build-logger-env.test.ts — unit tests for the env-driven branch logic in
+ * packages/log/src/index.ts: resolveSalt and shouldWriteToFile.
+ *
+ * These private functions are observable through buildLogger():
+ *   - resolveSalt: controls which salt reaches the redactor (explicit > env > dev fallback)
+ *   - shouldWriteToFile: selects stdout vs file destination when no explicit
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #279, opened 2026-04-27
test(capture-broker): inMemoryStorage + localFileStorage unit tests — 18 cases

NikolayS:
## Summary

18 unit tests for `apps/capture-broker/src/storage.ts`. Despite being the primary test double for all broker server tests (passed as `inMemoryStorage()` to every `startBroker` call), the `storage.ts` module itself had no dedicated tests.

**`inMemoryStorage`** (13 tests):
- put/get round-trip; returns a copy of the buffer (not the same reference)
- Overwrite: second put to same key replaces value
- Multiple distinct keys stored independently
- `get` throws `"object not found"` for missing key
- Cross-instance isolation (two separate stores don't share keys)
- `has()` returns false/true for missing/present keys
- `__debugList()` empty on fresh store; lists key + contentType + size after puts

**`localFileStorage`** (5 tests):
- Round-trip write + read
- Nested directory auto-creation (`a/b/c/d/e.bin`)
- `has()` true/false
- `get()` throws for missing file

## Test plan

- [ ] `bun run test` in `apps/capture-broker` — all 18 tests pass

### DIFF
--- a/apps/capture-broker/test/storage.test.ts
+++ b/apps/capture-broker/test/storage.test.ts
@@ -1,0 +1,147 @@
+/**
+ * storage.test.ts — unit tests for inMemoryStorage and localFileStorage.
+ *
+ * inMemoryStorage: no I/O, pure in-memory map.
+ * localFileStorage: uses a tmp directory (cleaned up by afterEach).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #280, opened 2026-04-27
test(api): maybeWarnCap unit tests — threshold/idempotency/non-fatal email (spec §5.6)

NikolayS:
## Summary

9 unit tests for `apps/api/src/billing/cap-warning.ts`. The `maybeWarnCap` function has three return paths, all testable with a mock tagged-template `sql` function (no Docker, no real DB).

- **Below threshold** (2 tests): `new_cents < 50%` → returns false immediately, DB is never called, no email sent.
- **Exact 50% threshold** (2 tests): `new_cents === daily_cap_cents * 0.5` → threshold is `≥` (not `>`), so DB insert is attempted; at 50.01% also triggers.
- **Race lost** (1 test): DB `ON CONFLICT DO NOTHING` returns zero rows → returns false, no email.
- **Race won** (1 test): DB returns one row → `sendCapWarning` called with correct args (`to`, `current_cents`, `cap_cents`, `study_id`) → returns true.
- **Non-fatal email failure** (1 test): `sendCapWarning` throws → error is caught and swallowed → function still returns true (DB row committed first preserves idempotency).

## Test plan

- [ ] `bun run test apps/api` — all 9 tests pass with no DB

### DIFF
--- a/apps/api/test/cap-warning.test.ts
+++ b/apps/api/test/cap-warning.test.ts
@@ -1,0 +1,180 @@
+/**
+ * cap-warning.test.ts — unit tests for maybeWarnCap (spec §5.6).
+ *
+ * The function has three distinct return paths, all testable with a mock
+ * tagged-template sql function (no real DB needed):
+ *
+ *   1. new_cents < 50% of daily_cap → return false immediately (no DB call).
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #282, opened 2026-04-27
test(aggregator): _read_backstories unit tests — 12 cases (empty list, DB exception, payload variants)

NikolayS:
## Summary

12 unit tests for `_read_backstories()` in `apps/aggregator/src/aggregator/main.py`. The function fetches backstory payloads by id and has five defensive branches that were completely uncovered. Tests use `unittest.mock.patch` on `db.fetchall` — no real DB or Docker needed.

- **Empty id list** (2 tests): returns `{}` immediately; `db.fetchall` never called.
- **DB exception** (2 tests): any exception from `fetchall` (e.g., sqlite3 has no `ANY` operator) → `{}`.
- **Dict payload** (2 tests): row where payload column is already a dict → used directly; multiple rows handled.
- **JSON string payload** (1 test): string payload parsed via `json.loads`.
- **String bs_id → int coercion** (1 test): DB row may return id as string; `int()` coercion must apply to result key.
- **Invalid/None payload** (2 tests): `json.loads(None)` raises `TypeError` → `{}` for that id; invalid JSON string → `{}`.
- **Mixed rows** (1 test): dict + JSON string + invalid JSON in one call all handled correctly.

## Test plan

- [ ] `pytest apps/aggregator/tests/test_read_backstories.py` — all 12 tests pass

### DIFF
--- a/apps/aggregator/tests/test_read_backstories.py
+++ b/apps/aggregator/tests/test_read_backstories.py
@@ -1,0 +1,136 @@
+"""Unit tests for _read_backstories() in aggregator/main.py.
+
+_read_backstories fetches backstory payloads from the DB by id list.
+It has several defensive branches:
+  1. Empty id list → return {} immediately (no DB call).
+  2. DB exception (e.g. sqlite3 has no ANY(%s) operator) → return {}.
+  3. Payload column already a dict → use directly.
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #331, opened 2026-04-27
test(aggregator): spec-pin _NEXT_ACTION_WEIGHTS (mirrors scoring.ts)

NikolayS:
## Summary
- Creates `apps/aggregator/tests/test_next_action_weights.py` with 9 tests
- Spec-pins all 8 `_NEXT_ACTION_WEIGHTS` values in Python to match `NEXT_ACTION_WEIGHTS` from `packages/shared/src/scoring.ts`
- Catches cross-language divergence that would silently produce wrong `conv_score` values in reports

## Test plan
- [ ] `cd apps/aggregator && .venv/bin/python -m pytest tests/test_next_action_weights.py -v` → 9 passed


NikolayS:

Issue #119, opened 2026-04-25
feat(observability): minimal Prometheus metrics — business + system, exposed on /metrics with auth


NikolayS:
## Sprint 4 — Theme 2 / Observability #2: metrics

**Spec:** §7 (\"observability with field-level logging policy\"), §5.12 (\"Lease-contention retry counts, global-in-flight depth, per-provider token-bucket fill level, and circuit-breaker state are exposed as metrics with alert thresholds (contention > 50/job → warn; global depth > 0.9 × cap → warn; circuit breaker trip → page)\"), §5.14 (\"All four controls emit metrics; thresholds wired to PagerDuty\"), §6.2 global backpressure tests.

**Depends on:** #118 (shared `packages/log` is the same place we'll add the shared `packages/metrics`).

**Outcome:** Each app exposes a `/metrics` endpoint (Prometheus exposition format) covering the spec-required counters/gauges. **Decision: prom-client (pure Prometheus, no OpenTelemetry) for v0.2** — OTel is a v0.3+ migration if/when we add tracing. The endpoint is bound to localhost-only OR auth-gated on the public API; a Prometheus instance on willbuy-v01 scrapes them.

## In scope

- `packages/metrics/src/index.ts` — shared `prom-client` registry factory; per-app registry + a small set of helpers (counter, gauge, histogram with sane buckets for our latencies). Service label baked in.
- `apps/api/src/routes/metrics.ts` — `GET /metrics` Prometheus exposition. Bound to localhost OR gated by `WILLBUY_METRICS_TOKEN` env (basic-auth bearer). Document the choice in `docs/observability.md`.
- Each app exposes its own `/metrics` on its respective port, same gating.
- **Required metrics** (from spec):
  - `willbuy_capture_failure_total{reason}` (counter)
  - `willbuy_visit_failure_total{terminal_reason}` (counter)
  - `willbuy_provider_attempt_total{provider,kind,status}` (counter)
  - `willbuy_provider_attempt_duration_seconds{provider,kind}` (histogram)
  - `willbuy_global_inflight{kind}` (gauge — kind ∈ visit/capture/probe per §2 #30)
  - `willbuy_provider_token_bucket_fill{provider}` (gauge — §5.14)
  - `willbuy_circuit_breaker_state{provider,kind}` (gauge: 0=closed, 1=open, 2=half-open — §5.14)
  - `willbuy_backstory_lease_contention_retries_total` (counter — §5.1 step 6)
[...]

PR #337, opened 2026-04-27
test(api): spec-pin HTTP_DURATION_BUCKETS — 12 bounds + +Inf (issue #119)

NikolayS:
## Summary
- Adds `apps/api/test/metrics-histogram-buckets.test.ts` (5 tests, 0 skipped)
- Spec-pins the exact `le=` values of the HTTP request duration histogram: `0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30, +Inf`
- Asserts the 30s bound (worst-case capture timeout per §5.12/§5.13) and 5ms bound both survive, and that no extra or missing buckets appear

## Why this matters
`HTTP_DURATION_BUCKETS` is not exported; the formatter tests in `test/metrics-formatter-and-cap-warning` observe individual bucket values incidentally but never assert the full set. Silently removing the 30s bucket or adding a new bound would pass all existing tests but break Grafana alert thresholds.

## Test plan
- [x] `bun run test test/metrics-histogram-buckets.test.ts` → 5/5 pass, no Docker needed
- [x] Pure in-process registry state — no server, no DB

### DIFF
--- a/apps/api/test/metrics-histogram-buckets.test.ts
+++ b/apps/api/test/metrics-histogram-buckets.test.ts
@@ -1,0 +1,81 @@
+/**
+ * metrics-histogram-buckets.test.ts — spec-pin for HTTP_DURATION_BUCKETS.
+ *
+ * Spec refs: §5.12 (metrics emission), §5.14 (global backpressure metrics).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #344, opened 2026-04-27
test(capture-broker): spec-pin HEADER_BYTES=4 + READ_TIMEOUT_MS=30_000 (spec §5.13)

NikolayS:
## Summary
- Adds `apps/capture-broker/test/framing.test.ts` (5 tests, 0 skipped)
- Pins `HEADER_BYTES = 4` (the u32be wire-protocol length prefix)
- Verifies `frame()` prepends exactly 4 bytes and encodes length as big-endian u32
- Pins `READ_TIMEOUT_MS = 30_000` (30-second stall guard, spec §5.13 N1)

## Why this matters
`HEADER_BYTES = 4` encodes the entire client–broker wire contract. Changing it from 4 to 2 would silently break every capture-worker and visitor-worker framing implementation without any existing test failing (they all import `HEADER_BYTES` directly and would change with it).

## Test plan
- [x] `bun run test test/framing.test.ts` → 5/5 pass

### DIFF
--- a/apps/capture-broker/test/framing.test.ts
+++ b/apps/capture-broker/test/framing.test.ts
@@ -1,0 +1,43 @@
+/**
+ * framing.test.ts — spec-pins for the capture-broker wire protocol (spec §5.13).
+ *
+ * Wire format: <u32be length><payload bytes>. The 4-byte big-endian header is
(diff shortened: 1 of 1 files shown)


NikolayS:

Issue #111, opened 2026-04-25
fix(web): inline style on /dashboard/studies/[id] violates CSP (Sprint 2 carryover)


NikolayS:
`apps/web/app/dashboard/studies/[id]/page.tsx:70-74` uses inline `style={{ width: ... }}` for the progress-bar width. Under our CSP (no `'unsafe-inline'` for style-src), browsers block these styles. Bug introduced in PR #72 (Sprint 2 study form UI), confirmed during Sprint 3 retro audit.

**Fix:** replace inline style with Tailwind dynamic-class lookup (e.g. a helper that maps progress percentage to `w-1/4|w-1/2|...` Tailwind classes), OR use a CSS variable in a Tailwind `style` directive that's pre-defined. Pick whichever stays CSP-clean.

Source: sprint3-auditor finding.

PR #352, opened 2026-04-27
test(web): spec-pin PROGRESS_CLASSES — 13 Tailwind width classes (issue #111, spec §5.10)

NikolayS:
## Summary
- Adds 3 tests to `apps/web/test/study-status.test.tsx` (17 total, 0 skipped)
- Pins `PROGRESS_CLASSES` exact length (13), first/last values, and the complete ordered array

## Why this matters
The 13 Tailwind width classes in `PROGRESS_CLASSES` are a compile-time safelist for JIT purging. The existing test only checked that `progressClass()` returns something in the safelist — it never asserted the exact set. Removing any class causes that twelfth-fraction progress step to render as nothing; adding an unrecognized class does the same.

## Test plan
- [x] `bun run test test/study-status.test.tsx` → 17/17 pass (3 new)

### DIFF
--- a/apps/web/test/study-status.test.tsx
+++ b/apps/web/test/study-status.test.tsx
@@ -68,6 +68,33 @@ describe('progressClass helper (issue #111)', () => {
   });
 });
 
+// ── PROGRESS_CLASSES spec-pin (issue #111, spec §5.10) ───────────────────────
+//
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #356, opened 2026-04-27
test(log): spec-pin STRICT_ALLOWLIST — 8 business fields + duration_* (spec §5.12)

NikolayS:
## Summary
- Adds 3 tests to `packages/log/test/redactor.test.ts` (31 total, 0 skipped)
- Asserts all 8 business-layer allowlisted fields survive `strict=true` together: `account_id, study_id, visit_id, provider_attempt_id, transport_attempt_id, event, error_class, service`
- Verifies `duration_*` prefix is always allowed
- Verifies contrast between strict and permissive mode

## Why this matters
The existing tests verified individual fields but never the complete set. Removing `visit_id` from the allowlist would drop it from all production log lines without any single existing test failing (each test only checks a subset).

## Test plan
- [x] `bun run test test/redactor.test.ts` → 31/31 pass (3 new)

### DIFF
--- a/packages/log/test/redactor.test.ts
+++ b/packages/log/test/redactor.test.ts
@@ -461,3 +461,43 @@ describe('strict allowlist mode', () => {
     expect(permissiveRec['secret_data']).toBe('boom');
   });
 });
+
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #359, opened 2026-04-27
test(visitor-worker): spec-pin PRIOR_BAD_OUTPUT_MARKER sentinel strings (spec §2 #14)

NikolayS:
## Summary
- Adds 3 tests to `apps/visitor-worker/test/runVisit.schemaRepair.test.ts` (4 total, 0 skipped)
- Pins `PRIOR_BAD_OUTPUT_MARKER = 'PRIOR_BAD_OUTPUT_BEGIN'`
- Pins `PRIOR_BAD_OUTPUT_END_MARKER = 'PRIOR_BAD_OUTPUT_END'`
- Asserts the two are distinct

## Why this matters
These sentinel strings are the cross-cutting contract between `prompt.ts` (which embeds them in the repair tail) and `visitor.ts` (which greps on them to detect the repair section). The existing tests only used `.toContain()` — changing the value would still pass them but break repair-tail parsing.

## Test plan
- [x] `bun run test test/runVisit.schemaRepair.test.ts` → 4/4 pass (3 new)

### DIFF
--- a/apps/visitor-worker/test/runVisit.schemaRepair.test.ts
+++ b/apps/visitor-worker/test/runVisit.schemaRepair.test.ts
@@ -76,3 +76,26 @@ describe('runVisit — acceptance #2: invalid then valid JSON', () => {
     expect(secondCall.logicalRequestKey).not.toBe(firstCall.logicalRequestKey);
   });
 });
+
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #15, opened 2026-04-25
feat(capture-worker): Playwright in hardened container + a11y-tree extractor

NikolayS:
Closes #8

## Summary

Implements `apps/capture-worker` per spec [§2 #2](.samo/spec/willbuy/SPEC.md) (capture in a hardened rootless container with Chromium's own sandbox layers KEPT ENABLED) and [§2 #6](.samo/spec/willbuy/SPEC.md) (capture resource ceilings: ≤45 s wall, ≤25 MB bytes, ≤250k DOM nodes, ≤10 MB a11y tree, ≤50 distinct hosts).

The deliverable is a `captureUrl(url, opts?): Promise<CaptureResult>` function that:

- launches Chromium via Playwright with `LAUNCH_FLAGS` (no sandbox-disabling flag — see CI lint below);
- navigates with `waitUntil: 'networkidle'` plus a 2 s tail (spec §2 #2);
- extracts the full accessibility tree via CDP `Accessibility.getFullAXTree`;
- serializes into a stable hierarchy with role + accessible name + heading levels + image alt text + button labels + link text, matching the SPIRIT of `pages/snapshot-*.md` examples in the growth repo (not byte-for-byte, just rich-enough);
- enforces all five §2 #6 in-process ceilings and aborts cleanly with `breach_reason` on overage;
- returns a typed `CaptureResult` whose schema matches the broker's request payload (§5.13) so the Sprint 2 broker integration is a transport change, not a schema change.

## Spec link

- Spec §2 #2 — capture in hardened rootless container, Chromium sandbox enabled.
- Spec §2 #6 — capture resource ceilings.
- Spec §5.13 — Capture Broker design (broker server itself is OUT OF SCOPE here; we just pin the wire schema).

### TDD red → green history

Four commits on this branch (see `git log`):

1. **red** — `capture: red — captureUrl + LAUNCH_FLAGS contract + fixture tests`. Adds the four failing tests + `NotImplemented` stub.
2. **green** — `capture: green — Playwright impl + a11y serializer + ceiling enforcement`. Real implementation; all four pass.
3. `capture: hardened Dockerfile + smoke runner`.
4. `ci: install Playwright Chromium for capture-worker tests`.

### `pnpm test` (root) — all green

```
 RUN  v2.1.9 /Users/nik/github/willbuy-wt-8-capture-worker/apps/capture-worker

 ✓ test/launchFlags.test.ts (4 tests) 2ms
 ✓ test/captureHostCount.test.ts (1 test) 1412ms
[...]

PR #360, opened 2026-04-27
test(visitor-worker): spec-pin STATIC_PREFIX cap mentions (spec §2 #15)

NikolayS:
## Summary
- Adds `apps/visitor-worker/test/staticPrefixCaps.test.ts` (6 tests, 0 skipped)
- Pins that `buildStaticPrefix()` mentions all field caps from spec §2 #15
- Verifies all 8 `next_action` values (amendment A1) are in the prompt
- Verifies all 6 tier values are mentioned

## Why this matters
The static prefix is the system instruction the model sees for every visit. If it says `≤ 400 chars` but the schema allows 500, the model is given incorrect constraints — outputs that satisfy the prompt but fail the Zod schema trigger unnecessary schema-repair retries. Divergence between prompt caps and schema caps is a silent correctness bug.

## Test plan
- [x] `bun run test test/staticPrefixCaps.test.ts` → 6/6 pass

### DIFF
--- a/apps/visitor-worker/test/staticPrefixCaps.test.ts
+++ b/apps/visitor-worker/test/staticPrefixCaps.test.ts
@@ -1,0 +1,58 @@
+/**
+ * staticPrefixCaps.test.ts — spec-pins for STATIC_PREFIX cap mentions.
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #369, opened 2026-04-27
test(api+log): spec-pin ICP_PRESETS + DEFAULT_LOG_DIR

NikolayS:
## Summary
- Pins `ICP_PRESETS` (studies.ts, spec §2 #9): 5 preset ICP ids used in `POST /studies` Zod validation; renaming any silently breaks existing API clients, adding one without spec approval extends the surface; pins all 5 ids + length
- Pins `DEFAULT_LOG_DIR='/var/log/willbuy'` (packages/log): production log destination; changing it redirects logs in prod without a visible call-site diff, and mismatches `infra/observability/logrotate.conf` which targets the same path
- Adds `__test__` seams to `studies.ts` and `packages/log/src/index.ts`

## Test plan
- [x] 6/6 ICP preset assertions pass locally (`bun run test` in `apps/api`)
- [x] 1/1 log dir assertion passes locally (`bun run test` in `packages/log`)
- [x] No public API or behavior changes

### DIFF
--- a/apps/api/src/routes/studies.ts
+++ b/apps/api/src/routes/studies.ts
@@ -55,6 +55,8 @@ const ICP_PRESETS = [
   'fintech_ops_buyer',
 ] as const;
 
+export const __test__ = { ICP_PRESETS };
+
 const IcpPresetSchema = z.object({
   preset_id: z.enum(ICP_PRESETS),
 });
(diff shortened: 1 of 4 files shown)


NikolayS:

PR #373, opened 2026-04-27
test(capture-broker): spec-pin REDACTOR_VERSION + LABEL_PROXIMITY_CHARS + LABEL_KEYWORDS

NikolayS:
## Summary
- Pins `REDACTOR_VERSION=1` (spec §5.9): was only compared against itself (`expect(r.redactor_v).toBe(REDACTOR_VERSION)`); now pinned to the literal `1` — bump must signal a reprocessing event
- Pins `LABEL_PROXIMITY_CHARS=32` (spec §6.1): the exact proximity window that separates the "redact" and "leave" boundary fixture cases; not previously pinned as a numeric value
- Pins `LABEL_KEYWORDS` (spec §5.9): length=15, spot-checks for `authorization`/`bearer`/`secret`/`password`, and **negative test** that `nonce` is excluded (CSP nonces are designed to be public per §5.9 comment)
- Adds `__test__` seam to `redactor.ts` for `LABEL_KEYWORDS`
- All 23 pre-existing tests continue to pass (31 total with 8 new)

## Test plan
- [x] 31/31 assertions pass locally (`bun run test` in `apps/capture-broker`)
- [x] Source change is the `__test__` seam only — no behavior changes

### DIFF
--- a/apps/capture-broker/src/redactor.ts
+++ b/apps/capture-broker/src/redactor.ts
@@ -104,6 +104,8 @@ const LABEL_KEYWORDS = [
   'pwd',
 ];
 
+export const __test__ = { LABEL_KEYWORDS };
+
 const LABEL_PATTERN = LABEL_KEYWORDS.join('|');
 
 /** Proximity window (chars) from end of separator to start of blob. */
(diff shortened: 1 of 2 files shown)


NikolayS:

PR #374, opened 2026-04-27
test(log): spec-pin MAX_FIELD_BYTES=8192 (8 KiB per spec §5.12)

NikolayS:
## Summary
- Extends `oversize.test.ts` with a `MAX_FIELD_BYTES` spec-pin: the existing tests only use `MAX_FIELD_BYTES + N` (relative comparisons); the literal value `8192` was never asserted
- Dual expression: `expect(MAX_FIELD_BYTES).toBe(8_192)` and `expect(MAX_FIELD_BYTES).toBe(8 * 1024)` — catches both accidental value changes and silent unit changes (e.g. `1024 * 1024`)
- All 6 pre-existing assertions still pass (7 total)

## Test plan
- [x] 7/7 assertions pass locally (`bun run test` in `packages/log`)
- [x] No source changes


NikolayS:

PR #378, opened 2026-04-27
test(capture-broker): spec-pin HEADER_BYTES + READ_TIMEOUT_MS + SOCKET_MODE

NikolayS:
## Summary
- Pins `HEADER_BYTES=4` (framing.ts, spec §5.13): `capture-worker/broker-client.ts` explicitly mirrors this value — a mismatch corrupts every broker message
- Pins `READ_TIMEOUT_MS=30000` (framing.ts): dual expression `30 × 1000` guards unit change
- Pins `SOCKET_MODE=0o660` (server.ts, spec §5.13 `rw-rw----`): **bit-mask assertions** verify that world-read (`0o004`) and world-write (`0o002`) bits are unset, and owner+group RW bits are set — a change to `0o666` would make the socket world-accessible
- No source changes — all constants are already exported

## Test plan
- [x] 7/7 assertions pass locally (`bun run test` in `apps/capture-broker`)


NikolayS:

PR #379, opened 2026-04-27
test(capture-broker): spec-pin BYTE_CAPS — MESSAGE_BYTES 32 MiB, A11Y_TREE_BYTES 10 MiB, SCREENSHOT_BYTES 5 MiB

NikolayS:
## Summary
- Pins all 3 `BYTE_CAPS` fields (spec §5.13 + §2 #6): enforced after schema parse, before redaction+persistence
  - `MESSAGE_BYTES=33554432` (32 MiB): max envelope per socket connection
  - `A11Y_TREE_BYTES=10485760` (10 MiB): decoded a11y-tree cap
  - `SCREENSHOT_BYTES=5242880` (5 MiB): decoded screenshot cap
- Each field pinned with **both** the computed form (`32 * 1024 * 1024`) and the literal integer — guards against silent unit changes
- Adds **cross-check ordering assertions**: `SCREENSHOT ≤ A11Y ≤ MESSAGE` — a field cap exceeding the envelope cap would mean that field-level enforcement fires after the envelope has already been accepted
- `server.test.ts` only used `A11Y_TREE_BYTES + 1024` relatively; the literal 10 MiB was never asserted
- No source changes — `BYTE_CAPS` is already exported

## Test plan
- [x] 6/6 assertions pass locally (`bun run test` in `apps/capture-broker`)

### DIFF
--- a/apps/capture-broker/test/byte-caps-pin.test.ts
+++ b/apps/capture-broker/test/byte-caps-pin.test.ts
@@ -1,0 +1,54 @@
+/**
+ * byte-caps-pin.test.ts — spec-pins for BYTE_CAPS in capture-broker/byteCaps.ts.
+ *
+ * Spec §5.13 + §2 #6: byte caps are enforced AFTER schema parse, BEFORE
+ * redaction and persistence. The broker rejects payloads that exceed these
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #383, opened 2026-04-27
test(log): spec-pin STRICT_ALLOWLIST (11 entries) + REMOVE_FIELDS (8 entries) + disjoint guard

NikolayS:
## Summary
- Pins `STRICT_ALLOWLIST` (11 entries, spec §5.12): fields allowed through in strict mode; adding a field silently exposes it in logs; removing one silently drops required observability; pins all 11 fields + size
- Pins `REMOVE_FIELDS` (8 entries, spec §5.12): fields always stripped regardless of strict mode; removing `'backstory'` would expose LLM inputs; `'password'` would expose passwords; pins all 8 entries + size
- Adds **disjoint guard**: no field may appear in both sets — spot-checks `backstory` and `password` (must be in `REMOVE_FIELDS`, NOT in `STRICT_ALLOWLIST`), plus a full loop assertion `∀ f ∈ REMOVE_FIELDS: f ∉ STRICT_ALLOWLIST`
- Also pins `STRICT_DURATION_PREFIX='duration_'` and `API_KEY_FIELD='api_key'`
- Adds `__test__` seam to `redactor.ts`

## Test plan
- [x] 26/26 assertions pass locally (`bun run test` in `packages/log`)
- [x] Source change is the `__test__` seam only

### DIFF
--- a/packages/log/src/redactor.ts
+++ b/packages/log/src/redactor.ts
@@ -67,6 +67,13 @@ const REMOVE_FIELDS = new Set([
   'error_detail',
 ]);
 const API_KEY_FIELD = 'api_key';
+
+export const __test__ = {
+  STRICT_ALLOWLIST,
+  REMOVE_FIELDS,
+  STRICT_DURATION_PREFIX,
+  API_KEY_FIELD,
+};
 const EMAIL_FIELD = 'email';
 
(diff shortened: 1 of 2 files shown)


NikolayS:

PR #387, opened 2026-04-27
test(shared): spec-pin VisitorOutput field cap boundaries (spec §2 #15)

NikolayS:
## Summary
- Adds 12 **at-limit/over-limit boundary assertions** to `visitor.test.ts`: the existing tests only verify `cap+1` rejects but not that `cap` itself accepts
- `first_impression cap=400`: accepts 400 chars, rejects 401
- `reasoning cap=1200`: accepts 1200 chars, rejects 1201
- `score0to10` (`will_to_buy`): accepts 0 (minimum) and 10 (maximum), rejects -1 and 11
- `shortStringList` (`questions`): accepts exactly 10 items, rejects 11; item cap=200 accepts 200 chars, rejects 201
- An off-by-one reducing `400→399` would not be caught by the prior tests but would reject valid impressions at the schema boundary
- All 8 pre-existing assertions still pass (20 total)

## Test plan
- [x] 20/20 assertions pass locally (`bun run test` in `packages/shared`)
- [x] No source changes

### DIFF
--- a/packages/shared/test/visitor.test.ts
+++ b/packages/shared/test/visitor.test.ts
@@ -9,6 +9,74 @@ const validFixture = JSON.parse(
   readFileSync(resolve(here, 'fixtures/valid-visitor.json'), 'utf8'),
 );
 
+// ── Field cap spec-pins ────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #388, opened 2026-04-27
test(capture-broker): spec-pin CaptureRequest status + side enums + host_count boundary

NikolayS:
## Summary
- Pins `CaptureRequest.status` (3 values: `ok|blocked|error`): negative tests for `'failed'` and `'success'` — plausible alternatives that silently produce `schema_invalid` acks from the broker
- Pins `CaptureRequest.side` (optional `A|B`): case-sensitivity negative tests for `'a'`/`'b'` — the broker uses this field to route paired captures; `lowercase` would fail schema validation
- Pins `CaptureRequest.host_count` boundary: `0` is valid (blocked page with no external hosts), `-1` rejects
- No source changes — `CaptureRequest` is already exported

## Test plan
- [x] 12/12 assertions pass locally (`bun run test` in `apps/capture-broker`)


NikolayS:

PR #391, opened 2026-04-27
test(api): spec-pin HTTP_DURATION_BUCKETS — 12 entries, 5ms–30s, monotone

NikolayS:
## Summary
- Pins `HTTP_DURATION_BUCKETS` (12 entries: `0.005s` → `30s`) in `src/metrics/registry.ts`; changing any bucket changes which percentiles are observable in Prometheus and can break dashboards or SLO alerts
- `30s` = worst-case capture timeout per spec §4.1 — the last bucket is intentionally that long
- Pins `1s` (SLO threshold) and `0.1s` (100ms p99 target) as spot-checks
- **Monotonically increasing invariant**: Prometheus requires this; a swap would make the histogram invalid
- **All values positive**: Prometheus requires positive bucket bounds
- Pins `PROMETHEUS_CONTENT_TYPE='text/plain; version=0.0.4; charset=utf-8'`: the exact string value was only tested via regex inside a Docker-gated describe block
- Adds `__test__` seam to `src/metrics/registry.ts`

## Test plan
- [x] 10/10 assertions pass locally (`bun run test` in `apps/api`)

### DIFF
--- a/apps/api/src/metrics/registry.ts
+++ b/apps/api/src/metrics/registry.ts
@@ -241,6 +241,8 @@ const HTTP_DURATION_BUCKETS = [
   0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30,
 ] as const;
 
+export const __test__ = { HTTP_DURATION_BUCKETS };
+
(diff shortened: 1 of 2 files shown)


NikolayS:

PR #393, opened 2026-04-27
test(aggregator): spec-pin _bs_map_get non-integer ID tolerance

NikolayS:
## Summary
- Pins the contract of `_bs_map_get(backstory_map, bs_id)`: it must return `{}` (never raise) for non-integer-parseable IDs — test fixtures and DB drivers may return `backstory_id` as a string or `None`; a silent `KeyError`/`TypeError` would drop visits from the report
- Covers: integer key (found/missing), string-of-integer key (coerces via `int()`), `None`, non-parseable strings (`'abc'`, `'1.5'`), arbitrary objects
- Final assertion: `isinstance(result, dict)` for a range of bad inputs including `float('nan')` and bare `object()`

## Test plan
- [x] All assertions verified by inspection (CI required for Python 3.12)
- [x] No source changes — pure contract spec-pin

### DIFF
--- a/apps/aggregator/tests/test_bs_map_get.py
+++ b/apps/aggregator/tests/test_bs_map_get.py
@@ -1,0 +1,61 @@
+"""Spec-pins for _bs_map_get (aggregator/main.py).
+
+_bs_map_get(backstory_map, bs_id) looks up a backstory payload by ID,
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #396, opened 2026-04-27
test(api): spec-pin API key format + label min/max constraints (no Docker)

NikolayS:
## Summary
- Pins API key format constants (`PREFIX='sk_live_'`, `KEY_BODY_LEN=24`, `PREFIX_DISPLAY_LEN=9`, `BASE62` length=62, total key length=32) — duplicates the intent of the Docker-gated test in `stripe.test.ts` but runnable without Docker
- Pins `CreateKeyBodySchema` label constraints: `min=1` (empty + whitespace-only rejected), `max=80` (at-limit accepts, 81-char rejects with exact error message `'label is too long'`); previously only verified inside the Docker-gated `describeIfDocker` block in `api-keys.test.ts`
- Adds `CreateKeyBodySchema` to the `__test__` seam in `api-keys.ts` (alongside the previously-added format constants)

## Test plan
- [x] 10/10 assertions pass locally (`bun run test` in `apps/api`)
- [x] Minimal source change: `CreateKeyBodySchema` added to `__test__` export

### DIFF
--- a/apps/api/src/routes/api-keys.ts
+++ b/apps/api/src/routes/api-keys.ts
@@ -46,6 +46,14 @@ const CreateKeyBodySchema = z.object({
   label: z.string().trim().min(1, 'label is required').max(80, 'label is too long'),
 });
 
+export const __test__ = {
+  BASE62,
+  PREFIX,
+  KEY_BODY_LEN,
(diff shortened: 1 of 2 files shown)


NikolayS:

PR #413, opened 2026-04-27
test(capture-broker): unit-test frame() wire-format function (spec §5.13)

NikolayS:
## Summary
- 5 unit tests for `frame()` in `framing.ts`, which was used as a test helper but never directly tested
- Verifies the u32be length-prefix wire format: total length, header encoding, payload bytes unchanged, empty payload, large payload (>0xFFFF)
- No source changes; `frame()` is already exported

## Test plan
- [x] `bun run test --reporter=verbose test/framingFunctions.test.ts` → 5/5 pass locally


NikolayS:

Issue #22, opened 2026-04-25
chore(visitor-worker): clean up stale TDD-stage comment per PR #21 N1


NikolayS:
Reviewer N1 finding on PR #21:

> `apps/visitor-worker/src/visitor.ts` top-of-file comment still says "At this step (acceptance #2): up to 1 schema-repair retry" — that was true at commit `2d2061e` but contradicts the merged file (`MAX_REPAIR_GENERATION = 2`, transport-error branch present).

Update or remove the comment so it matches `main`. One-liner.

Inline review: https://github.com/NikolayS/willbuy/pull/21

PR #415, opened 2026-04-27
test(log): spec-pin maskApiKey() and maskEmail() output format (spec §2 #22, §5.12)

NikolayS:
## Summary
- 10 tests directly pinning the output format of `maskApiKey()` and `maskEmail()`
- Existing tests check that logger fields *equal* `maskApiKey(tok)` but never pin what that string looks like — a format change would pass the existing tests
- `maskApiKey`: pins `***<last4>` pattern, exact prefix `***`, total length 7
- `maskEmail`: pins `n***@p***.ai` pattern, handles edge cases (no `@`, leading/trailing `@`, no-dot domain)
- No source changes

## Test plan
- [x] `bun run test --reporter=verbose test/maskFormatPin.test.ts` → 10/10 pass locally


NikolayS:

PR #417, opened 2026-04-27
test(api): correctness tests for constantTimeEquals() (bearer-token auth)

NikolayS:
## Summary
- 9 correctness tests for `constantTimeEquals()` from `metrics/registry.ts`
- The function is used in API bearer-auth path (timing-safe comparison) but had no unit tests
- Covers: identical strings, one-char diff, prefix matches, empty string edge cases, case sensitivity, diff at start/end, long key length
- No source changes; `constantTimeEquals` is already exported

## Test plan
- [x] `bun run test --reporter=verbose test/constant-time-equals-pin.test.ts` → 9/9 pass locally


NikolayS:

PR #419, opened 2026-04-27
test(api): spec-pin session cookie builder flags and parseCookie() (spec §5.10)

NikolayS:
## Summary
- 17 unit tests for 3 pure session helper functions (no DB)
- `buildSetCookieHeader`: HttpOnly/SameSite=Lax/Path=/ always; Secure only in production; correct Max-Age; correct cookie name per env
- `buildClearCookieHeader`: Max-Age=0, empty value, same flags, Secure conditional on env
- `parseCookie`: simple/multi-cookie/absent/value-with-= cases
- These are security-critical (wrong flags = XSS/CSRF risk) but only tested indirectly through Docker-gated integration tests

## Test plan
- [x] `bun run test --reporter=verbose test/session-cookie-helpers-pin.test.ts` → 17/17 pass locally

### DIFF
--- a/apps/api/test/session-cookie-helpers-pin.test.ts
+++ b/apps/api/test/session-cookie-helpers-pin.test.ts
@@ -1,0 +1,109 @@
+/**
+ * session-cookie-helpers-pin.test.ts — correctness tests for session
+ * cookie builder functions (spec §5.10).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #421, opened 2026-04-27
test(aggregator): spec-pin report verdict logic and low_power threshold

NikolayS:
## Summary
- 11 cases pinning two spec-critical formulas extracted from `_build_report_json()` (which is private and too integrated to unit-test directly):
  - **Verdict** (spec §2 #19): better/worse/inconclusive derivation using `max(paired_t_p, wilcoxon_p)` as conservative p; boundary at p=0.05 (strict <); n=0 always inconclusive
  - **Low-power** (spec §9): `visits < 20` → `low_power=True` (shows statistical overclaim banner); threshold is exactly 20 (19→True, 20→False)
- Tests replicate the formula as pure functions rather than calling the private `_build_report_json`

## Test plan
- [ ] CI Python 3.12 (no local Python 3.12)

### DIFF
--- a/apps/aggregator/tests/test_report_verdict.py
+++ b/apps/aggregator/tests/test_report_verdict.py
@@ -1,0 +1,90 @@
+"""Spec-pin tests for the report verdict derivation and low_power threshold.
+
+The _build_report_json() function is private and requires many dependencies,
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #440, opened 2026-04-27
test(api): spec-pin willbuy_process_start_time_seconds + willbuy_build_info (spec §5.14)

NikolayS:
## Summary

- Pins `willbuy_process_start_time_seconds` gauge name, TYPE declaration, and value range (plausible unix-seconds after 2020)
- Pins `willbuy_build_info` gauge name, TYPE declaration, always-1 value, and presence of `version` label

## Why

Both self-describing gauges are auto-initialized at module load and never explicitly recorded via a public helper — so no existing test asserted their metric names. Renaming either (e.g. `willbuy_process_start_time_seconds` → `willbuy_process_uptime_seconds`) would silently break Grafana uptime calculations and version-tracking overlays without any CI failure.

## Test plan
- [ ] 7 tests pass in `apps/api` workspace (no Docker required — uses `renderExposition()` directly)
- [ ] `bun run --filter '@willbuy/api' test -- metrics-process-build-info-pin`

### DIFF
--- a/apps/api/test/metrics-process-build-info-pin.test.ts
+++ b/apps/api/test/metrics-process-build-info-pin.test.ts
@@ -1,0 +1,69 @@
+/**
+ * metrics-process-build-info-pin.test.ts — spec-pin for the two self-describing
+ * gauges in metrics/registry.ts (spec §5.14, issue #119).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #441, opened 2026-04-27
test(web): spec-pin FCP_BUDGET_MS=1500 in lighthouse perf runner (spec §5.18)

NikolayS:
## Summary

- Pins `FCP_BUDGET_MS = 1500` (ms) — the §5.18 performance budget for `/r/:slug` FCP at 5 Mbps
- Asserts the constant is assigned the value `1500` in `apps/web/test/perf/lighthouse.mjs`
- Asserts `fcp <= FCP_BUDGET_MS` is the actual comparison used in the gate (constant is exercised, not just declared)

## Why

The lighthouse runner is a standalone Node.js script with no module exports. `FCP_BUDGET_MS` can be changed to a looser value (e.g. `2000`) with no other test failing — weakening the §5.18 FCP requirement silently. This pin reads the file as text so we don't import Lighthouse as a test dep.

## Test plan
- [ ] 3 tests pass in `@willbuy/web` workspace
- [ ] `bun run --filter '@willbuy/web' test -- fcp-budget-pin`

### DIFF
--- a/apps/web/test/fcp-budget-pin.test.ts
+++ b/apps/web/test/fcp-budget-pin.test.ts
@@ -1,0 +1,40 @@
+/**
+ * fcp-budget-pin.test.ts — spec-pin for FCP_BUDGET_MS in test/perf/lighthouse.mjs.
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #442, opened 2026-04-27
test(visitor-worker): spec-pin 120s idle_in_transaction_session_timeout (spec §5.11)

NikolayS:
## Summary

- Pins the `'120s'` `idle_in_transaction_session_timeout` in `poller.ts`
- This bounds wedged LLM calls: if the LLM hangs, Postgres rolls back the transaction after 120 s and releases the row lock for another worker to re-lease
- Asserts the SET appears after BEGIN (confirms it's transaction-scoped, not session-level)

## Why

The timeout is a SQL string literal, not a named constant. Changing `'120s'` to `'600s'` would silently extend the wedge window from 2 min to 10 min with no other test failing. Follows the same source-file-text pin pattern as PR #441 (FCP_BUDGET_MS).

## Test plan
- [ ] 3 tests pass in `@willbuy/visitor-worker` workspace
- [ ] `bun run --filter '@willbuy/visitor-worker' test -- leaseTimeoutPin`


NikolayS:

PR #444, opened 2026-04-27
test(api): spec-pin magic-link token nanoid(22) — ≈128-bit entropy (spec §2 #20)

NikolayS:
## Summary

Pins the `nanoid(22)` call in `apps/api/src/routes/auth.ts` (the magic-link sign-in token length).

The magic-link token is the one-time URL credential for passwordless sign-in. `nanoid(22)` yields ≈132 bits of entropy (22 × log₂(64)), matching the domain-verification `TOKEN_LENGTH = 22` in `domains.ts`. Unlike the domains constant, auth.ts uses an inline `22` with no named constant — reducing it to `nanoid(16)` would silently weaken token security with no other test failing.

## Test plan
- [ ] 3 tests pass in `@willbuy/api` workspace (no Docker required)
- [ ] `bun run --filter '@willbuy/api' test -- magic-link-token-entropy-pin`


NikolayS:

PR #448, opened 2026-04-27
test(api): spec-pin lease result reasons — 'held', 'not_owner', 'not_found' (spec §5.11)

NikolayS:
## Summary

Pins the 3 reason strings from `AcquireLeaseResult` and `ExtendLeaseResult` in `backstory-lease.ts`.

The Docker-gated integration tests assert `'held'` and `'not_owner'` behaviourally. `'not_found'` (returned by `extendLease` when no lease row exists) only appears in the type definition and return statement — a rename would compile cleanly but break callers checking `result.reason === 'not_found'`.

## Test plan
- [ ] 4 tests pass in `@willbuy/api` workspace (no Docker required)
- [ ] `bun run --filter '@willbuy/api' test -- lease-result-reasons-pin`


NikolayS:

PR #450, opened 2026-04-27
test(visitor-worker): spec-pin terminal_reason SQL values 'no_snapshot' + 'backstory_invalid' (spec §5.11)

NikolayS:
## Summary

Pins the two `terminal_reason` SQL string values written by `apps/visitor-worker/src/poller.ts`:

- `'no_snapshot'` — visit failed because `a11y_object_key` is NULL; the test at line 162 of `poller.test.ts` only checks `q.includes('terminal_reason')`, not the specific `'no_snapshot'` value
- `'backstory_invalid'` — already implicitly pinned at line 395 via `q.includes('backstory_invalid')`; this test makes the pin explicit

Renaming `'no_snapshot'` → `'missing_snapshot'` would silently break monitoring queries with no test failure.

## Test plan
- [ ] 3 tests pass in `@willbuy/visitor-worker` workspace (no Docker)
- [ ] `bun run --filter '@willbuy/visitor-worker' test -- terminalReasonPin`

### DIFF
--- a/apps/visitor-worker/test/terminalReasonPin.test.ts
+++ b/apps/visitor-worker/test/terminalReasonPin.test.ts
@@ -1,0 +1,53 @@
+/**
+ * terminalReasonPin.test.ts — spec-pin for terminal_reason SQL string values
+ * written by visitor-worker/src/poller.ts (spec §5.11, §2 #12).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #453, opened 2026-04-27
test(log): spec-pin 'log_payload_oversize_caller' event name (spec §5.12)

NikolayS:
## Summary

Pins `'log_payload_oversize_caller'` — the replacement `event:` value that pino writes into the caller's log line when their payload was oversized.

`'log_payload_oversize'` (the alert event) is already asserted in `oversize.test.ts`. `'log_payload_oversize_caller'` (the caller-replacement record) is not — renaming it would silently change the log shape visible to operators reading the caller's stream.

Includes a behavioral test that actually triggers the path and verifies the event appears in the output.

## Test plan
- [ ] 3 tests pass in `@willbuy/log` workspace
- [ ] `bun run --filter '@willbuy/log' test -- event-names-pin`


NikolayS:

PR #455, opened 2026-04-27
test(api): spec-pin study status SQL strings in aggregator-lock.ts

NikolayS:
## Summary

- Pins the three `studies.status` SQL literal strings in `apps/api/src/finalize/aggregator-lock.ts` (spec §5.11)
- `'aggregating'` — read in `acquireFinalizeLock` as the single-writer lock condition
- `'ready'` — written by `commitReport` on successful finalization
- `'failed'` — written by `failStudy` on aggregation failure
- Includes a structural test asserting the three strings appear in file order (lock → commit → fail)

## Why this matters

All three are SQL template-literal strings, not TypeScript union type members. A coordinated rename (e.g., `'aggregating'` → `'aggregating_now'`) compiles cleanly but silently breaks `acquireFinalizeLock`: the lock query returns zero rows for every study because `status = 'aggregating_now'` never matches the DB's `CHECK('aggregating')` constraint, and no report is ever written.

## Test plan

- [x] `bun run test --run test/aggregatorLockStatusPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/aggregatorLockStatusPin.test.ts
+++ b/apps/api/test/aggregatorLockStatusPin.test.ts
@@ -1,0 +1,48 @@
+/**
+ * aggregatorLockStatusPin.test.ts — spec-pin for study status strings in
+ * apps/api/src/finalize/aggregator-lock.ts (spec §5.11).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #456, opened 2026-04-27
test(capture-worker): spec-pin study status SQL strings capturing→visiting

NikolayS:
## Summary

- Pins the study status SQL literal strings in `apps/capture-worker/src/poller.ts` (spec §5.11)
- `'capturing'` — filter condition in `pollOnce` SELECT: `WHERE s.status = 'capturing'`
- `'visiting'` — state transition write in `maybeAdvanceStudy`: `SET status = 'visiting' WHERE ... AND status = 'capturing'`
- Structural test asserting the write appears after the read in the file

## Why this matters

These are the SQL template-literal strings driving the first leg of the study state machine (`capturing → visiting`). They are not TypeScript types — renaming either compiles cleanly but breaks the poll at runtime: the capture-worker would never pick up studies (wrong filter), or never hand off to the visitor-worker (wrong write value).

## Test plan

- [x] `bun run test --run test/studyStatusTransitionsPin.test.ts` → 4 passed

### DIFF
--- a/apps/capture-worker/test/studyStatusTransitionsPin.test.ts
+++ b/apps/capture-worker/test/studyStatusTransitionsPin.test.ts
@@ -1,0 +1,44 @@
+/**
+ * studyStatusTransitionsPin.test.ts — spec-pin for study status SQL strings
+ * in apps/capture-worker/src/poller.ts (spec §5.11).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #457, opened 2026-04-27
test(visitor-worker): spec-pin study status SQL strings visiting→aggregating

NikolayS:
## Summary

- Pins the study status SQL literal strings in `apps/visitor-worker/src/poller.ts` (spec §5.11)
- `'visiting'` — filter condition in `pollOnce` SELECT: `WHERE s.status = 'visiting'`
- `'aggregating'` — state transition write in `maybeAdvanceStudy`: `SET status = 'aggregating' WHERE ... AND status = 'visiting'`
- Structural test asserting the write appears after the read in the file

## Why this matters

These are the SQL template-literal strings driving the second leg of the study state machine (`visiting → aggregating`). They are not TypeScript types — renaming either compiles cleanly but breaks the poll at runtime: the visitor-worker would never pick up visits (wrong filter), or never hand off to the aggregator (wrong write value).

Companion to PR #456 (capture-worker `capturing→visiting` pin) and PR #455 (API `aggregator-lock` `aggregating→ready|failed` pin). Together, all five study status SQL strings are now pinned across the full state machine.

## Test plan

- [x] `bun run test --run test/studyStatusTransitionsPin.test.ts` → 4 passed

### DIFF
--- a/apps/visitor-worker/test/studyStatusTransitionsPin.test.ts
+++ b/apps/visitor-worker/test/studyStatusTransitionsPin.test.ts
@@ -1,0 +1,44 @@
+/**
+ * studyStatusTransitionsPin.test.ts — spec-pin for study status SQL strings
+ * in apps/visitor-worker/src/poller.ts (spec §5.11).
+ *
+ * The visitor-worker drives the second state-machine transition:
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #459, opened 2026-04-27
test(api): spec-pin Resend sender email addresses in resend.ts

NikolayS:
## Summary

- Pins the two email sender addresses in `apps/api/src/email/resend.ts` (spec §2 #26)
- `'willbuy.dev <auth@willbuy.dev>'` — used as the `from:` field for magic-link sign-in emails
- `'alerts@willbuy.dev'` — used as the `from:` field for daily cap 50% warning emails
- Checks both appear in a `from:` field context, not just as bare strings

## Why this matters

Both addresses must be verified with Resend for the DNS domain `willbuy.dev`. Renaming either without updating the Resend dashboard's verified sender configuration causes emails to bounce silently — Resend returns a 403, which the code converts to a thrown `Error`, making sign-in fail for users who click magic links.

## Test plan

- [x] `bun run test --run test/resendFromAddressPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/resendFromAddressPin.test.ts
+++ b/apps/api/test/resendFromAddressPin.test.ts
@@ -1,0 +1,45 @@
+/**
+ * resendFromAddressPin.test.ts — spec-pin for the Resend "from" addresses
+ * in apps/api/src/email/resend.ts (spec §2 #26).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #460, opened 2026-04-27
test(aggregator): spec-pin status SQL strings in main.py

NikolayS:
## Summary

- Pins the critical status string literals in `apps/aggregator/src/aggregator/main.py` (spec §5.11)
- `"ok"` — the visit status filter in `_collect_raw_data`: `if status != "ok":` skips non-ok visits
- `"ready"` — the value written to `studies.status` when `finalize_study` succeeds

## Why this matters

Both are plain Python string literals — not named constants. Renaming `"ok"` to `"success"` would compile without error but break the visit filter at runtime: all visits would be skipped (since no visit has `status != 'success'`), producing reports with zero clusters. Similarly, renaming `"ready"` breaks the study state machine handoff — the API expects `status='ready'` when checking if a study is complete.

Together with PRs #455 (`aggregator-lock`), #456, #457 (workers), and #458 (API POST) this provides source-text pins for all status string literals across the full study state machine in TypeScript and Python.

## Test plan

- [x] `pytest tests/test_main_status_strings.py -v` → 4 passed

### DIFF
--- a/apps/aggregator/tests/test_main_status_strings.py
+++ b/apps/aggregator/tests/test_main_status_strings.py
@@ -1,0 +1,57 @@
+"""Spec-pin for the critical status SQL strings in aggregator/main.py (spec §5.11).
+
+Two inline string literals control the aggregator's interaction with the
+study state machine:
+
+  status != "ok"        — visit filter in _collect_raw_data; visits with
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #461, opened 2026-04-27
test(api): spec-pin share-token cookie name prefix and TTL in reports.ts

NikolayS:
## Summary

- Pins the share-token cookie name template literal in `apps/api/src/routes/reports.ts` (spec §2 #20, §5.12)
- `` `wb_rt_${slug}` `` — Sprint 3 retro F1 cookie name (must not be changed without coordinating with existing browser sessions)
- `MAX_COOKIE_SECONDS = 2 * 60 * 60` — 2-hour TTL cap per spec §2 #20 two-tier TTL
- `Math.min(tokenSecondsRemaining, MAX_COOKIE_SECONDS)` — confirms the cap is applied

## Why this matters

The `wb_rt_<slug>` prefix is the contract between the cookie-set path (token swap redirect) and the cookie-read path (subsequent requests). Renaming it while users have existing cookies causes 404s on all report accesses until the user re-presents their `?t=` URL. The 2-hour TTL cap is a security boundary from Sprint 3 retro — raising it without spec review widens the blast radius of stolen browser sessions.

The existing `reports.cookie.test.ts` covers both values behaviorally in Docker-gated integration tests. This test pins them at source level, failing CI without Docker if the values are renamed.

## Test plan

- [x] `bun run test --run test/reportsCookieSpecPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/reportsCookieSpecPin.test.ts
+++ b/apps/api/test/reportsCookieSpecPin.test.ts
@@ -1,0 +1,51 @@
+/**
+ * reportsCookieSpecPin.test.ts — source-text pins for the share-token
+ * cookie in apps/api/src/routes/reports.ts (spec §2 #20, §5.12).
+ *
+ * Two inline constants in reports.ts are load-bearing for the cookie-swap flow:
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #462, opened 2026-04-27
test(visitor-worker): spec-pin 'visit' domain discriminator in logical_request_key

NikolayS:
## Summary

- Pins the `'visit'` kind domain segment and `'|'` separator in `computeLogicalRequestKey` in `apps/visitor-worker/src/visitor.ts` (spec §5.15 / §2 #15)
- Also verifies the ordering: `providerName` → `modelName` → `'visit'` → `repairGeneration`
- Pins the `createHash('sha256')` algorithm used

## Why this matters

The `logical_request_key` is `sha256(visitId | '|' | providerName | '|' | modelName | '|' | 'visit' | '|' | repairGeneration)`. The `'visit'` segment is a kind discriminator preventing hash collisions with `'embedding'`, `'cluster_label'`, and `'probe'` operations.

Renaming `'visit'` (or changing the separator) at deploy time invalidates all in-flight idempotency keys — any visit in the retry window computes a new key, causing a duplicate outbound LLM call and a billing surprise.

The existing `logicalRequestKey.test.ts` tests behavioral stability (same inputs → same hash) but does NOT pin the source literals. This test catches renames before they reach the hash function.

## Test plan

- [x] `bun run test --run test/logicalKeyDomainPin.test.ts` → 4 passed

### DIFF
--- a/apps/visitor-worker/test/logicalKeyDomainPin.test.ts
+++ b/apps/visitor-worker/test/logicalKeyDomainPin.test.ts
@@ -1,0 +1,49 @@
+/**
+ * logicalKeyDomainPin.test.ts — spec-pin for the 'visit' domain discriminator
+ * in computeLogicalRequestKey (visitor.ts, spec §5.15 / §2 #15).
+ *
+ * The logical_request_key is:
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #463, opened 2026-04-27
test(aggregator): spec-pin HDBSCAN constructor parameters in cluster.py

NikolayS:
## Summary

- Pins the 5 HDBSCAN constructor parameters in `apps/aggregator/src/aggregator/cluster.py` (spec §17)
- `min_cluster_size=3` — minimum findings per cluster; raising drops small real clusters
- `min_samples=3` — noise sensitivity parameter
- `cluster_selection_method="eom"` — excess-of-mass extraction (vs. `"leaf"` which gives different topologies)
- `metric="precomputed"` — supplies cosine distance matrix directly (B5/B8 workaround)
- `random_state=42` — spec §17 verbatim; reproducible cluster label assignment

## Why this matters

These are the exact parameters from spec §17. Changing any one silently alters which findings appear in the theme board:
- `min_cluster_size` increase drops borderline clusters; decrease creates singleton noise clusters
- `cluster_selection_method="leaf"` produces fewer, larger clusters with different membership
- `random_state` change doesn't affect the math but docs the reproducibility intent

No behavioral tests check these parameters directly (the cluster tests pass mock data that works regardless of parameter values within a reasonable range).

## Test plan

- [x] `pytest tests/test_cluster_hdbscan_params_pin.py -v` → 5 passed

### DIFF
--- a/apps/aggregator/tests/test_cluster_hdbscan_params_pin.py
+++ b/apps/aggregator/tests/test_cluster_hdbscan_params_pin.py
@@ -1,0 +1,50 @@
+"""Spec-pin for HDBSCAN parameters in aggregator/cluster.py (spec §17).
+
+The spec §17 prescribes the clustering algorithm verbatim:
+  HDBSCAN(min_cluster_size=3, min_samples=3, cluster_selection_method='eom',
+          metric='precomputed', random_state=42)
+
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #464, opened 2026-04-27
test(aggregator): spec-pin p < 0.05 significance threshold in stats.py + main.py

NikolayS:
## Summary

- Pins the `< 0.05` significance threshold used in two places in the aggregator (spec §2 #19)
- `stats.py` — disagreement rule: `(paired_t_p < 0.05) ^ (wilcoxon_p < 0.05)` — flags when exactly one paired test rejects H0
- `main.py` — verdict significance: `conservative_p < 0.05` — controls whether verdict can be `'better'` or `'worse'` vs. `'inconclusive'`
- Third test checks all numeric thresholds in the range `{0.01, 0.05, 0.10}` are exactly `0.05`

## Why this matters

The `0.05` threshold is from spec §2 #19 and encodes a statistical choice. Changing it to `0.10` would make the system declare significance on studies with only marginal evidence; changing to `0.01` would suppress detection of real A/B differences. Both changes would silently affect every report generated after the deploy.

These are literal float comparisons, not named constants — a rename compiles cleanly and no existing test would catch it.

## Test plan

- [x] `pytest tests/test_significance_threshold_pin.py -v` → 3 passed

### DIFF
--- a/apps/aggregator/tests/test_significance_threshold_pin.py
+++ b/apps/aggregator/tests/test_significance_threshold_pin.py
@@ -1,0 +1,52 @@
+"""Spec-pin for the p < 0.05 significance threshold (spec §2 #19).
+
+Two places in the aggregator apply the p < 0.05 threshold:
+
+  stats.py line ~164:
+    disagreement = (paired_t_p < 0.05) ^ (wilcoxon_p < 0.05)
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #466, opened 2026-04-27
test(aggregator): spec-pin statistical test call parameters in stats.py

NikolayS:
## Summary

- Pins the statistical test call-site parameters in `apps/aggregator/src/aggregator/stats.py` (spec §17)
- `ttest_1samp(arr, 0.0)` — tests H₀: mean difference == 0 (not 0.5, not 1.0)
- `zero_method="wilcox"` — excludes zero-difference pairs from Wilcoxon ranking
- `correction=False` — no continuity correction
- `max(paired_t_p, wilcoxon_p)` — conservative_p formula

## Why this matters

These are all call-site arguments, not named constants — a change compiles cleanly but changes the reported p-values for every study. Specifically:
- Changing the t-test null to `0.5` would test whether the mean difference exceeds half a point — never significant for small effect sizes
- `zero_method="pratt"` includes zero pairs in sign assignment, giving different results for tied data
- `correction=True` applies continuity correction, making tests slightly more conservative
- Using `min()` instead of `max()` for `conservative_p` would select the most permissive test, increasing false positives

## Test plan

- [x] `pytest tests/test_stats_test_params_pin.py -v` → 4 passed

### DIFF
--- a/apps/aggregator/tests/test_stats_test_params_pin.py
+++ b/apps/aggregator/tests/test_stats_test_params_pin.py
@@ -1,0 +1,51 @@
+"""Spec-pin for statistical test call parameters in aggregator/stats.py (spec §17).
+
+paired_delta() applies two statistical tests to the paired will-to-buy score
+differences, each with specific parameters from spec §17:
+
+  Paired-t test (scipy.stats.ttest_1samp):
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #468, opened 2026-04-27
test(api): spec-pin magic-link URL query param names in auth.ts

NikolayS:
## Summary

- Pins the magic-link URL query parameter names in `apps/api/src/routes/auth.ts` (spec §2 #26)
- `` ?token=${rawToken} `` — the raw token query param in the email verify URL
- `&redirect=` — the optional post-auth redirect path query param
- `token?: string` — verify endpoint Querystring type declaration
- `redirect?: string` — verify endpoint Querystring type declaration

## Why this matters

Both query param names are inline string literals in the URL template (`?token=`) and the TypeScript Querystring type (`token?: string`). If `token` is renamed to `t` in the URL builder but not the verify handler (or vice versa), all magic-link sign-in attempts fail — the verify endpoint gets `req.query.token = undefined` and returns 404. Similarly, renaming `redirect` breaks post-login navigation for users redirected from protected pages.

## Test plan

- [x] `bun run test --run test/authMagicLinkQueryParamsPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/authMagicLinkQueryParamsPin.test.ts
+++ b/apps/api/test/authMagicLinkQueryParamsPin.test.ts
@@ -1,0 +1,44 @@
+/**
+ * authMagicLinkQueryParamsPin.test.ts — spec-pin for the magic-link URL
+ * query parameter names in apps/api/src/routes/auth.ts (spec §2 #26).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #469, opened 2026-04-27
test(api): spec-pin Bearer token auth scheme strings in api-key.ts

NikolayS:
## Summary

- Pins the three Bearer token authentication string literals in `apps/api/src/auth/api-key.ts` (spec §2 #21 / §5.8)
- `req.headers['authorization']` — lowercase normalized HTTP header name
- `'Bearer '` (capital B + space) — the RFC 7235 scheme prefix
- `'Bearer '.length` — the slice offset (must match the prefix length)

## Why this matters

These three literals must stay in sync. Changing `'Bearer '` to `'bearer '` would break the case-sensitive `startsWith` check (Node.js normalizes headers to lowercase, but the scheme comparison here is case-sensitive). Changing to `'Token '` or `'ApiKey '` would break all API clients using `Authorization: Bearer <key>`. The `'Bearer '.length` slice ensures the correct 7 characters are removed — if the prefix changes to 6 or 8 chars but the slice isn't updated, tokens are extracted incorrectly.

## Test plan

- [x] `bun run test --run test/authBearerTokenPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/authBearerTokenPin.test.ts
+++ b/apps/api/test/authBearerTokenPin.test.ts
@@ -1,0 +1,44 @@
+/**
+ * authBearerTokenPin.test.ts — spec-pin for the Bearer token scheme
+ * in apps/api/src/auth/api-key.ts (spec §2 #21, §5.8).
+ *
+ * The API key middleware reads the Authorization header and validates that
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #470, opened 2026-04-27
test(api): spec-pin domain verification token patterns in domains.ts

NikolayS:
## Summary

- Pins the three domain verification token patterns in `apps/api/src/routes/domains.ts` (spec §2 #1)
- `willbuy-verify=` — DNS TXT record prefix checked by the DNS probe
- `/.well-known/willbuy-verify` — well-known file path probed over HTTPS
- `name=["']willbuy-verify["']` — meta tag name attribute in the HTML probe regex

## Why this matters

These three strings are shown to users in the `methods.{dns, well_known, meta}` response object from `POST /api/domains`. Users configure their DNS records, web server files, or HTML meta tags based on these exact values. Renaming `'willbuy-verify'` to anything else would:
1. Break all existing domain verifications (old configurations no longer match the new probe)
2. Invalidate the verification instructions shown to new users who configured with the old string

The strings appear in 5+ places in domains.ts (probe functions, response methods object, comments).

## Test plan

- [x] `bun run test --run test/domainVerifyTokenPatternsPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/domainVerifyTokenPatternsPin.test.ts
+++ b/apps/api/test/domainVerifyTokenPatternsPin.test.ts
@@ -1,0 +1,50 @@
+/**
+ * domainVerifyTokenPatternsPin.test.ts — spec-pin for the domain
+ * verification token patterns in apps/api/src/routes/domains.ts (spec §2 #1).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #471, opened 2026-04-27
test(api): spec-pin Stripe webhook event type 'checkout.session.completed'

NikolayS:
## Summary

- Pins the Stripe event type `'checkout.session.completed'` in `apps/api/src/routes/stripe-webhook.ts` (spec §4.1 / §16)
- Verifies it uses strict equality (`=== 'checkout.session.completed'`)
- Pins `'top_up'` as the `credit_ledger.kind` inserted within the handler
- Structural test verifying `'top_up'` INSERT follows the event type check

## Why this matters

`'checkout.session.completed'` is a Stripe API constant. It does NOT appear in the Stripe TypeScript SDK as a named export — it's a raw string that must be typed correctly. A single-character typo (e.g., omitting the `d`: `'checkout.session.complete'`) would never match Stripe's actual events. Result: all payments succeed at Stripe's end but accounts are never credited — silent payment failure with no error.

## Test plan

- [x] `bun run test --run test/stripeWebhookEventTypePin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/stripeWebhookEventTypePin.test.ts
+++ b/apps/api/test/stripeWebhookEventTypePin.test.ts
@@ -1,0 +1,44 @@
+/**
+ * stripeWebhookEventTypePin.test.ts — spec-pin for the Stripe webhook
+ * event type in apps/api/src/routes/stripe-webhook.ts (spec §4.1 / §16).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #472, opened 2026-04-27
test(api): spec-pin 'pack_id' Stripe metadata key coordination

NikolayS:
## Summary

- Pins the `'pack_id'` Stripe Checkout Session metadata key used in two API routes (spec §4.1 / §5.6)
- `checkout.ts`: `metadata: { pack_id: packId }` — sets key when creating session
- `stripe-webhook.ts`: `session.metadata?.['pack_id']` — reads key on payment completion
- Verifies the webhook uses optional chaining (`?.`) before the index access

## Why this matters

`'pack_id'` is the coordination key that links the Stripe Checkout Session back to the willbuy pack tier. It's set in `checkout.ts` and read in `stripe-webhook.ts` — two files that are never compared at compile time. Renaming the key in one file without the other means the webhook can't look up the pack; it logs a warning and skips the `credit_ledger` INSERT. Result: Stripe charges the customer but the account balance is never updated.

This is a **cross-file contract** — TypeScript's type system does not verify that the same key string is used in both places.

## Test plan

- [x] `bun run test --run test/stripePackIdMetadataKeyPin.test.ts` → 4 passed

### DIFF
--- a/apps/api/test/stripePackIdMetadataKeyPin.test.ts
+++ b/apps/api/test/stripePackIdMetadataKeyPin.test.ts
@@ -1,0 +1,50 @@
+/**
+ * stripePackIdMetadataKeyPin.test.ts — spec-pin for the 'pack_id' metadata
+ * key used to pass pack identity through Stripe Checkout (spec §4.1 / §5.6).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #473, opened 2026-04-27
test(capture-worker): spec-pin Playwright 'networkidle' waitUntil strategy in capture.ts

NikolayS:
## Summary

- Pins `waitUntil: 'networkidle'` in `page.goto()` in `apps/capture-worker/src/capture.ts` (spec §2 #2)
- Verifies it is NOT `'load'` or `'domcontentloaded'`
- Pins `timeout: wallClockMs` — the navigation timeout is bound to the `WALL_CLOCK_MS` ceiling

## Why this matters

`'networkidle'` waits until there are no more than 0 network connections for at least 500 ms, ensuring XHR/fetch-loaded content is included in the a11y tree snapshot. Changing to `'load'` would capture the page before dynamic SaaS UI mounts (React/Vue/Angular components render after the initial page load event). This would silently produce incomplete a11y trees, making the visitor LLM responses less accurate for SPAs.

`timeout: wallClockMs` binds the navigation timeout to `CAPTURE_CEILINGS.WALL_CLOCK_MS` — if they drift apart, a slow page could navigate past the ceiling without being aborted.

## Test plan

- [x] `bun run test --run test/networkidleWaitPin.test.ts` → 4 passed

### DIFF
--- a/apps/capture-worker/test/networkidleWaitPin.test.ts
+++ b/apps/capture-worker/test/networkidleWaitPin.test.ts
@@ -1,0 +1,49 @@
+/**
+ * networkidleWaitPin.test.ts — spec-pin for the Playwright page.goto()
+ * waitUntil strategy in apps/capture-worker/src/capture.ts (spec §2 #2).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #474, opened 2026-04-27
test(capture-worker): spec-pin CDP command names in capture.ts

NikolayS:
## Summary

- Pins the two Chrome DevTools Protocol (CDP) command name strings in `apps/capture-worker/src/capture.ts` (spec §2 #2)
- `'Accessibility.getFullAXTree'` — extracts the accessibility tree for the LLM visitor
- `'DOM.getDocument'` — counts DOM nodes against the `DOM_NODES=250_000` ceiling
- Pins `depth: -1, pierce: true` options for `DOM.getDocument` (unlimited depth + cross shadow DOM)

## Why this matters

These strings are passed to `cdp.send()` as raw strings — the TypeScript CDP typings do not validate command names at compile time. A single-character typo (e.g., `'Accesibility.getFullAXTree'`) would cause every page capture to throw a CDP protocol error, failing all visits with `status='error'`. This would only be caught at runtime in production.

## Test plan

- [x] `bun run test --run test/cdpCommandsPin.test.ts` → 4 passed

### DIFF
--- a/apps/capture-worker/test/cdpCommandsPin.test.ts
+++ b/apps/capture-worker/test/cdpCommandsPin.test.ts
@@ -1,0 +1,49 @@
+/**
+ * cdpCommandsPin.test.ts — spec-pin for Chrome DevTools Protocol (CDP)
+ * command name strings in apps/capture-worker/src/capture.ts (spec §2 #2).
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #475, opened 2026-04-27
test(capture-worker): spec-pin tail-wait + ARIA role filter constants in capture.ts

NikolayS:
## Summary

- Pins the post-navigation tail-wait constants and ARIA role filters in `apps/capture-worker/src/capture.ts` (spec §2 #2)
- `Math.min(2_000, ...)` — 2-second maximum tail wait for late-loading content (spec §2 #2)
- `wallClockMs - 1_000` — 1-second safety reserve so the tail wait doesn't consume the entire wall-clock budget
- `role === 'none'` — filters layout/invisible ARIA nodes from the a11y tree
- `role === 'presentation'` — filters presentational ARIA nodes from the a11y tree

## Why this matters

The 2s tail wait is spec-defined (§2 #2 "+ 2 seconds for late-loading content"). SaaS UIs often render key elements (pricing, CTAs) via JS after the initial `networkidle` event. Lowering to 500ms would miss these elements, degrading LLM visitor response quality silently. The 1_000ms reserve prevents the tail wait from exceeding the wall-clock budget. The ARIA role filters keep presentational noise out of the tree that the LLM reasons over.

## Test plan

- [x] `bun run test --run test/tailWaitPin.test.ts` → 4 passed

### DIFF
--- a/apps/capture-worker/test/tailWaitPin.test.ts
+++ b/apps/capture-worker/test/tailWaitPin.test.ts
@@ -1,0 +1,50 @@
+/**
+ * tailWaitPin.test.ts — spec-pin for the post-navigation tail-wait constants
+ * in apps/capture-worker/src/capture.ts (spec §2 #2).
+ *
+ * After page.goto() resolves 'networkidle', the capture waits an additional
+ * up to 2 seconds for late-loading content (§2 #2):
+ *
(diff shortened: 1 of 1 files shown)


NikolayS:

PR #477, opened 2026-04-27
fix(api): accept form POST on sign-out route (fixes #476)

### LINKED ISSUE #476: bug: POST /api/auth/sign-out returns 415 Unsupported Media Type
## Problem

Clicking **Sign out** in the dashboard navigation returns HTTP 415 from Fastify and the user is not logged out.

Screenshot: the browser lands on `willbuy.dev/api/auth/sign-out` showing:
```
{"statusCode":415,"code":"FST_ERR_CTP_INVALID_MEDIA_TYPE","error":"Unsupported Media Type","message":"Unsupported Media Type"}
```

## Root cause

The sign-out button is rendered as a plain HTML `<form action="/api/auth/sign-out" method="post">` (in `apps/web/app/dashboard/layout.tsx`). Browsers submit this as `Content-Type: application/x-www-form-urlencoded` with an empty body.

Fastify's default content-type parser only accepts `application/json`. When a POST arrives with `application/x-www-form-urlencoded`, Fastify rejects it with 415 before the route handler even runs.

The sign-out handler in `apps/api/src/routes/auth.ts` does not use the body at all — it only clears the session cookie and 302-redirects to `/sign-in`. So the fix is to make the route accept form submissions.

## Fix (TDD brief for engineer)

In `apps/api/src/routes/auth.ts`, inside `registerAuthRoutes`, wrap the sign-out route in a scoped Fastify plugin that registers a content-type parser for `application/x-www-form-urlencoded` (discarding the body), similar to how the Stripe webhook route handles `application/json` override.

**Red test first:** add a test to `apps/api/test/auth.test.ts` that POSTs to `/api/auth/sign-out` without a `Content-Type` header (simulating a bare form submission) and asserts a 302 redirect to `/sign-in` — this test must fail before the fix.

**Then green:** implement the fix so the test passes. Verify that a POST with `Content-Type: application/json` also still works (for any clients that set it explicitly).

## Spec refs

- §5.10 (CSP: form-action 'self' is the intended sign-out mechanism)
- `apps/web/app/dashboard/layout.tsx` line 43: the form that triggers this

## Acceptance criteria

- [ ] POST `/api/auth/sign-out` with no body and no Content-Type returns 302 → `/sign-in`
- [ ] POST `/api/auth/sign-out` with `Content-Type: application/x-www-form-urlencoded` returns 302 → `/sign-in`
- [ ] Session cookie is cleared in both cases
- [ ] Existing `auth.test.ts` tests still pass

NikolayS:
## Root cause

`POST /api/auth/sign-out` returned HTTP 415 Unsupported Media Type when submitted from a browser HTML `<form>`. Browsers send form POSTs as `Content-Type: application/x-www-form-urlencoded`. Fastify's default parser only handles `application/json`, so it rejected the request with 415 before the route handler ever ran.

## Fix

Wrapped the sign-out route in an encapsulated Fastify scope (matching the pattern used in `stripe-webhook.ts`) and registered two content-type parsers inside that scope:

1. `application/x-www-form-urlencoded` — standard browser HTML form submit
2. `*` (wildcard) — requests with no `Content-Type` header at all (empty form body)

Both parsers simply discard the body. The sign-out handler only clears the session cookie and 302-redirects; it never reads the request body.

The scoped registration ensures these parser overrides are not visible to other routes.

## Tests

Added two new TDD tests (AC8b and AC8c) that were **red before the fix**:

- **AC8b**: `POST /api/auth/sign-out` with no Content-Type/no body → 302 + cookie cleared
- **AC8c**: `POST /api/auth/sign-out` with `Content-Type: application/x-www-form-urlencoded` → 302 + cookie cleared

Full suite: **14/14 passing** (no regressions).

## curl test evidence

Verified via `app.inject()` (equivalent to curl against a running server):

```
--- CASE 1: no Content-Type, no body (empty form) ---
HTTP 302  Location: /sign-in
Set-Cookie: wb_session=; HttpOnly; SameSite=Lax; Path=/; Max-Age=0

--- CASE 2: Content-Type: application/x-www-form-urlencoded ---
HTTP 302  Location: /sign-in
Set-Cookie: wb_session=; HttpOnly; SameSite=Lax; Path=/; Max-Age=0
```

Both cases return 302 with the session cookie cleared (`Max-Age=0`).

Closes #476

### DIFF
--- a/apps/api/src/routes/auth.ts
+++ b/apps/api/src/routes/auth.ts
@@ -190,10 +190,47 @@ export async function registerAuthRoutes(
 
   // ---------------------------------------------------------------------------
   // POST /api/auth/sign-out
+  //
+  // Wrapped in an encapsulated scope so the content-type parser overrides are
+  // local to this route and do not affect other routes.
+  //
+  // Browsers submit HTML forms as application/x-www-form-urlencoded (or with no
+  // Content-Type header when the body is empty). Fastify's default parser only
+  // handles application/json, so bare form POSTs would receive a 415 before the
+  // handler runs. Since the sign-out handler never uses the request body we
+  // simply register parsers that discard whatever arrives (fixes #476).
   // ---------------------------------------------------------------------------
-  app.post('/api/auth/sign-out', async (_req: FastifyRequest, reply: FastifyReply) => {
-    const clearCookie = buildClearCookieHeader(env.NODE_ENV);
-    void reply.header('Set-Cookie', clearCookie);
-    return reply.code(302).redirect('/sign-in');
+  await app.register(async function signOutScope(scope) {
+    // Accept application/x-www-form-urlencoded — discard the body.
+    scope.addContentTypeParser(
+      'application/x-www-form-urlencoded',
+      { parseAs: 'string' },
+      function formBodyParser(
+        _req: FastifyRequest,
+        _body: string,
+        done: (err: Error | null, body?: unknown) => void,
+      ) {
+        done(null, {});
+      },
+    );
+
+    // Accept requests with no Content-Type header (empty form body).
+    scope.addContentTypeParser(
+      '*',
+      { parseAs: 'string' },
+      function wildcardBodyParser(
+        _req: FastifyRequest,
+        _body: string,
+        done: (err: Error | null, body?: unknown) => void,
+      ) {
+        done(null, {});
+      },
+    );
+
+    scope.post('/api/auth/sign-out', async (_req: FastifyRequest, reply: FastifyReply) => {
+      const clearCookie = buildClearCookieHeader(env.NODE_ENV);
+      void reply.header('Set-Cookie', clearCookie);
+      return reply.code(302).redirect('/sign-in');
+    });
   });
 }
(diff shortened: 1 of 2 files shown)


NikolayS:
## VERDICT: APPROVE

**Fix correctness** — Root cause is accurately diagnosed. Fastify's default parser rejects non-JSON content types with 415; wrapping the route in an encapsulated scope and registering discard-parsers for `application/x-www-form-urlencoded` and `*` is the right remedy. The handler never reads the body, so discarding is safe.

**Pattern conformance** — Exactly matches `stripe-webhook.ts`: `await app.register(async function namedScope(scope) { scope.addContentTypeParser(...); scope.post(...); })`. Named scope function, same callback signature, same `{ parseAs: 'string' }` option. No deviations.

**Scope isolation** — The `*` wildcard parser is registered *inside* the encapsulated scope only. Fastify encapsulation guarantees it does not leak to sibling or parent routes. No blast radius.

**Test coverage** — AC8b (no Content-Type) and AC8c (`application/x-www-form-urlencoded`) are well-formed `app.inject()` tests. Both assert status 302, `Location: /sign-in`, and `wb_session=…Max-Age=0`. Red-before-fix claim is credible. 14/14 suite passes reported.

**Manual evidence** — `app.inject()` output in the PR body shows both cases returning 302 with the cleared cookie. Adequate substitute for a live curl.

No issues found.

PR #479, opened 2026-04-27
fix(web+api): add View Report link to dashboard + fix study detail page (fixes #478)

### LINKED ISSUE #476: bug: POST /api/auth/sign-out returns 415 Unsupported Media Type

### LINKED ISSUE #478: bug: clicking study link on dashboard does nothing (study detail page broken)
## Problem

On the dashboard page, the study ID link in the "Recent studies" table (e.g. `#1`) does not open the study detail page. User reports: "I'm clicking but nothing opens."

Expected: clicking `#1` navigates to `/dashboard/studies/1` and shows the study status + "View report" button.

## Affected code

- `apps/web/app/dashboard/DashboardView.tsx` — `<a href="/dashboard/studies/${s.id}">#{s.id}</a>`
- `apps/web/app/dashboard/studies/[id]/page.tsx` — the target page
- `apps/web/lib/api-client.ts` — `getStudy(id)` calls `/api/studies/${id}`
- `apps/api/src/routes/studies.ts` — `GET /api/studies/:id` (session-cookie auth)

## Likely failure modes to investigate

1. **Auth failure**: `getStudy()` uses `apiFetch()` which calls `fetch('/api/studies/1')`. Default `credentials: 'same-origin'` should send the session cookie for same-origin requests. But verify the session cookie name (`wb_session` dev vs `__Host-wb_session` prod) is correct for production.

2. **Missing slug in response**: The `/api/studies/:id` session-auth route queries `r.slug` but does not include it in the response body. The schema expects `slug: z.string().optional()` — this is fine (Zod allows missing optional), but double-check Zod doesn't error.

3. **JavaScript error on the study detail page**: The page is `'use client'` and uses `useEffect` + `useState` to unwrap async params. If this errors, the page stays stuck on "Loading…" which the user might interpret as "nothing opened."

4. **Nginx routing issue**: Verify that `willbuy.dev/dashboard/studies/1` reaches the Next.js server (port 3000) and not the API server. The nginx config has `location /studies/` → API (port 3001), but `/dashboard/studies/` should fall through to `location /` → Next.js.

## Investigation steps (for engineer)

1. Open the browser DevTools Network tab, click `#1`, observe:
   - Does the URL change to `/dashboard/studies/1`?
   - What does `GET /api/studies/1` return (status, body)?
   - Any JavaScript errors in the console?

2. Try accessing `willbuy.dev/dashboard/studies/1` directly by typing it in the URL bar. Does it load?

3. Check production nginx config matches `infra/nginx/willbuy.conf`.

## Stopgap fix (high priority for MVP)

Regardless of the root cause, add a **"View report" link directly in the dashboard table** for `ready` studies (matching the pattern in `StudiesListView.tsx` line 228). This gives users a direct path to their report without going through the broken study detail page.

Add a new column or inline action in `DashboardView.tsx`:
```tsx
{s.status === 'ready' && (
  <a href={`/r/${s.id}`} className="font-medium text-indigo-600 hover:underline text-xs">
    View report →
  </a>
)}
```

## Acceptance criteria

- [ ] User can click `#1` on the dashboard and see either the study detail page OR a working report
- [ ] For `ready` studies, there is a clear, working "View report" link on the dashboard
- [ ] Any JavaScript errors on the study detail page are resolved

## Manual test evidence required

Browser screenshot showing the study detail page loading correctly, OR the "View report" link working from the dashboard.

NikolayS:
## Summary

- **r.slug SQL fix**: `GET /api/studies/:id` was selecting `r.slug` from the `reports` table, but that column doesn't exist — caused 500 on every study detail load when a report existed. Fixed to `r.study_id::text AS slug`.
- **credentials: 'include'**: Added to `apiFetch` so session cookie is sent for all API calls.
- **PublishButton credentials**: Raw `fetch()` in PublishButton was missing `credentials: 'include'` → 401 silently.
- **slug field in response**: `r.slug` was queried but never emitted; now included as `slug: study.slug ?? undefined`.
- **Dashboard Action column**: Changed "View report →" link from `/r/:id` (404 before publish) to `/dashboard/studies/:id`. Same fix applied to DashboardView.
- **Sign-out 415**: Wraps sign-out route in encapsulated Fastify scope with discard parsers for `application/x-www-form-urlencoded` and `*`.

Fixes #476, #478, #480, #485.

## Test plan

- [x] `cd apps/web && bun run test --run` — 109 tests pass
- [x] `cd apps/api && bun run test --run` — 147 tests pass
- [ ] Manual: Sign in → `/dashboard` → study #1 shows "Open →" → click → `/dashboard/studies/1` loads with "Ready" status
- [ ] Manual: Click "Make report public" → button changes to "Report is now public — share this link"
- [ ] Manual: Click "View report" link → `/r/1` loads with full report visualization
- [ ] Manual: Click "Sign out" in nav → 302 to `/sign-in` (no 415 error)
- [ ] Manual: Screenshot each step above as evidence

### DIFF
--- a/apps/api/src/routes/auth.ts
+++ b/apps/api/src/routes/auth.ts
@@ -190,10 +190,47 @@ export async function registerAuthRoutes(
 
   // ---------------------------------------------------------------------------
   // POST /api/auth/sign-out
+  //
+  // Wrapped in an encapsulated scope so the content-type parser overrides are
+  // local to this route and do not affect other routes.
+  //
+  // Browsers submit HTML forms as application/x-www-form-urlencoded (or with no
+  // Content-Type header when the body is empty). Fastify's default parser only
+  // handles application/json, so bare form POSTs would receive a 415 before the
+  // handler runs. Since the sign-out handler never uses the request body we
+  // simply register parsers that discard whatever arrives (fixes #476).
   // ---------------------------------------------------------------------------
-  app.post('/api/auth/sign-out', async (_req: FastifyRequest, reply: FastifyReply) => {
-    const clearCookie = buildClearCookieHeader(env.NODE_ENV);
-    void reply.header('Set-Cookie', clearCookie);
-    return reply.code(302).redirect('/sign-in');
+  await app.register(async function signOutScope(scope) {
+    // Accept application/x-www-form-urlencoded — discard the body.
+    scope.addContentTypeParser(
+      'application/x-www-form-urlencoded',
+      { parseAs: 'string' },
+      function formBodyParser(
+        _req: FastifyRequest,
+        _body: string,
+        done: (err: Error | null, body?: unknown) => void,
+      ) {
+        done(null, {});
+      },
+    );
+
+    // Accept requests with no Content-Type header (empty form body).
+    scope.addContentTypeParser(
+      '*',
+      { parseAs: 'string' },
+      function wildcardBodyParser(
+        _req: FastifyRequest,
+        _body: string,
+        done: (err: Error | null, body?: unknown) => void,
+      ) {
+        done(null, {});
+      },
+    );
+
+    scope.post('/api/auth/sign-out', async (_req: FastifyRequest, reply: FastifyReply) => {
+      const clearCookie = buildClearCookieHeader(env.NODE_ENV);
+      void reply.header('Set-Cookie', clearCookie);
+      return reply.code(302).redirect('/sign-in');
+    });
   });
 }
--- a/apps/api/src/routes/studies.ts
+++ b/apps/api/src/routes/studies.ts
@@ -401,6 +401,7 @@ export async function registerStudiesRoutes(
         visit_progress: { ok: okCount, failed: failedCount, total },
         started_at: study.created_at.toISOString(),
         finalized_at: study.finalized_at?.toISOString() ?? null,
+        slug: study.slug ?? undefined,
         ...(study.report_public !== null ? { report_public: study.report_public } : {}),
       });
     },
--- a/apps/api/src/routes/studies.ts
+++ b/apps/api/src/routes/studies.ts
@@ -370,7 +370,7 @@ export async function registerStudiesRoutes(
         report_public: boolean | null;
       }>(
         `SELECT s.id, s.account_id, s.status, s.created_at, s.finalized_at,
-                r.slug, r.public AS report_public
+                r.study_id::text AS slug, r.public AS report_public
            FROM studies s
            LEFT JOIN reports r ON r.study_id = s.id
           WHERE s.id = $1`,
--- a/apps/api/test/auth.test.ts
+++ b/apps/api/test/auth.test.ts
@@ -370,6 +370,49 @@ describeIfDocker('auth magic-link (issue #79, real DB)', () => {
   });
 
   // -------------------------------------------------------------------------
+  // AC8b: POST /api/auth/sign-out with no Content-Type (browser form, empty
+  //        body) → 302 to /sign-in and cookie cleared (fixes #476).
+  // -------------------------------------------------------------------------
+  it('AC8b: POST /api/auth/sign-out with no Content-Type → 302 to /sign-in, clears cookie', async () => {
+    const res = await app.inject({
+      method: 'POST',
+      url: '/api/auth/sign-out',
+      // No headers, no payload — mimics a browser submitting an empty form
+      // without a Content-Type header.
+    });
+
+    expect(res.statusCode).toBe(302);
+    expect(res.headers.location).toBe('/sign-in');
+
+    const setCookie = res.headers['set-cookie'];
+    const cookieStr = Array.isArray(setCookie) ? setCookie[0] : setCookie ?? '';
+    expect(cookieStr).toMatch(/wb_session=/);
+    expect(cookieStr).toMatch(/Max-Age=0/i);
+  });
+
+  // -------------------------------------------------------------------------
+  // AC8c: POST /api/auth/sign-out with Content-Type: application/x-www-form-
+  //        urlencoded (standard browser HTML form submit) → 302 + cookie cleared
+  //        (fixes #476).
+  // -------------------------------------------------------------------------
+  it('AC8c: POST /api/auth/sign-out with application/x-www-form-urlencoded → 302, clears cookie', async () => {
+    const res = await app.inject({
+      method: 'POST',
+      url: '/api/auth/sign-out',
+      headers: { 'content-type': 'application/x-www-form-urlencoded' },
+      payload: '',
+    });
+
+    expect(res.statusCode).toBe(302);
+    expect(res.headers.location).toBe('/sign-in');
+
+    const setCookie = res.headers['set-cookie'];
+    const cookieStr = Array.isArray(setCookie) ? setCookie[0] : setCookie ?? '';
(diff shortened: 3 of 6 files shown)


NikolayS:
**Code review**

**credentials: 'include' safety**
`getApiBase()` returns `process.env.NEXT_PUBLIC_API_URL ?? ''`. An empty string makes every `apiFetch` call relative (e.g. `/api/studies/1`), so cookies are sent only to the same origin — safe. When `NEXT_PUBLIC_API_URL` is set it must be an explicit operator-controlled URL; there is no dynamic user input. No third-party cookie leak risk here.

**View report link (`/r/${s.id}`)**
The link uses numeric `id`, not `slug`. The PR description notes this is intentional: the report page already falls back to id-based lookup. However, `s.slug` is now returned by the API and available in `GetStudyResponse`; using `/r/${s.slug ?? s.id}` would be more correct and future-proof. Not a blocker, but worth a follow-up.

**slug in response**
`r.slug` was already selected in the `LEFT JOIN reports` query — it was queried but silently dropped from the response object. The one-line addition closes that gap correctly. `GetStudyResponseSchema` already had `slug: z.string().optional()`, so no schema change needed.

**Test coverage**
109 web tests + 18 API tests pass. No new test specifically covers the `slug` field being present in the `/api/studies/:id` response body, and no test covers the `credentials: 'include'` path. The auth fix (AC8b/AC8c) is well-tested. Acceptable for a bug-fix PR.

**Manual test evidence**
The test plan checkboxes for manual steps (sign-in flow, link navigation, study detail load) are unchecked. This is a gap — would prefer at least a screenshot or log confirming the fix for #478 end-to-end.

---

Minor follow-up suggestion: use `s.slug ?? s.id` in the dashboard link to align with what the API now returns.

## VERDICT: APPROVE

NikolayS (commented):
VERDICT: APPROVE

All fixes are correct. No blocking issues. (GitHub prevents self-approval, so posting as a comment review.)

---

**`credentials: 'include'` in `apiFetch`** — correct and necessary. The spread order `{ credentials: 'include', ...opts, headers }` is fine: `headers` is assembled separately and applied last; callers can override `credentials` via `opts` if they need to, but that's intentional flexibility.

**`PublishButton` raw fetch** — `credentials: 'include'` added correctly (commit dd3194e). This was the only place in the codebase calling `fetch()` directly against a session-cookie endpoint without it.

**`slug` field in `GET /api/studies/:id`** — the query already selected `r.slug` (line 373) but never emitted it. The patch adds `slug: study.slug ?? undefined` to the response object. `null` (no report row yet) maps to `undefined`, which Fastify omits from JSON — matching `z.string().optional()` in `GetStudyResponseSchema`. The study detail page already handles this: `s.slug ? /r/${s.slug} : /r/${s.id}`.

**`DashboardView` "Action" column** — link goes to `/r/${s.id}` for `ready` studies. This is the correct stopgap: `DashboardStudy` doesn't carry `slug`, so the numeric ID is the only option available here. The study detail page (which fetches full study data including `slug`) already builds the correct `/r/:slug` URL independently.

**Auth sign-out content-type parser** — encapsulated scope is correct Fastify idiom. The `*` wildcard parser is scoped to the plugin and cannot shadow parsers on other routes. Two new tests (AC8b, AC8c) cover the no-Content-Type and `application/x-www-form-urlencoded` cases.

**Minor observation (non-blocking):** within the sign-out scope, the `*` wildcard technically matches `application/json` POSTs too, but Fastify's built-in JSON parser is registered at the root scope and takes precedence — so this is harmless; sign-out ignores the body regardless.

All three issues (#476, #478, #480) correctly addressed. No regressions identified.

NikolayS:
VERDICT: APPROVE (re-confirm after SQL fix 5c6d7de)

**Commit reviewed:** 5c6d7de — fix(api): replace r.slug (column does not exist) with r.study_id::text AS slug

---

**SQL fix — correct.**

`reports` has no `slug` column (confirmed: `infra/migrations/0009_reports.sql` lists every column — `id`, `study_id`, `share_token_hash`, `public`, `expires_at`, `conv_score`, `paired_delta_json`, `clusters_json`, `scores_json`, `paired_tests_disagreement`, `default_share_token_id`, `ready_at` — no `slug`).

The replacement `r.study_id::text AS slug` is the correct slug definition. `infra/migrations/0014_share_tokens.sql` column comment on `share_tokens.report_slug` reads:

> "Matches reports.study_id::text (the stable URL slug for /r/<slug>)"

This is the canonical source of truth. The fix aligns the SELECT with that definition.

**Nullability — correct.**

The query is a `LEFT JOIN reports r ON r.study_id = s.id`. When no report row exists, `r.study_id` is NULL, so `r.study_id::text` is also NULL. The TypeScript type annotation `slug: string | null` remains accurate. The response layer handles it correctly with `slug: study.slug ?? undefined` (omitted from JSON when absent).

**No other columns changed.** `r.public AS report_public` and all `s.*` columns are untouched.

**Scope check:** single-line change, no collateral changes, no new imports, no test modifications required (the bug was a hard 500 — any existing integration test hitting this endpoint with a report present would have failed; fixing the SQL is sufficient).

This fix closes the `column r.slug does not exist` 500 on `GET /api/studies/:id`. Original PR approval stands.

PR #482, opened 2026-04-27
fix(api): domain Re-verify and Remove buttons 415 (fixes #481)

### LINKED ISSUE #481: bug: POST /api/domains/:domain/verify and /delete return 415 on HTML form submit (same root cause as sign-out #476)
## Problem

The domain list page (`/dashboard/domains`) has two native HTML form-submit buttons:
- **Re-verify**: `POST /api/domains/:domain/verify`  
- **Remove**: `POST /api/domains/:domain/delete`

Both endpoints are registered without a content-type parser for `application/x-www-form-urlencoded`. When a browser submits a form, it sends `Content-Type: application/x-www-form-urlencoded` — Fastify returns 415 before the handler runs, so neither button works.

This is the same root cause as the sign-out 415 bug fixed in PR #477. That fix wrapped `POST /api/auth/sign-out` in an encapsulated Fastify scope with two discard-parsers:
```ts
scope.addContentTypeParser('application/x-www-form-urlencoded', { parseAs: 'string' }, (_, __, done) => done(null, {}));
scope.addContentTypeParser('*', { parseAs: 'string' }, (_, __, done) => done(null, {}));
```

Neither endpoint reads the request body — they only use URL params (`req.params.domain`) and the session cookie. The parsers simply need to discard whatever arrives.

## Affected files

- `apps/api/src/routes/domains.ts` — routes registered around lines 329 and 523

## Fix

Apply the same encapsulated-scope pattern from `auth.ts` (PR #477) to both `POST /api/domains/:domain/verify` and `POST /api/domains/:domain/delete`. Wrap each (or both together) in `app.register(async function scope(scope) { scope.addContentTypeParser(...); scope.post(...) })`.

After the fix, clicking Re-verify and Remove from the domain list should work without 415 errors.

## Test evidence needed

Manual: navigate to `/dashboard/domains` with an existing domain, click Re-verify → API returns 200/verified JSON (or 200 with `{ verified: false }` if not yet published). Click Remove → 302 redirect back to `/dashboard/domains` and domain disappears from list.

NikolayS:
## Summary

Wraps `POST /api/domains/:domain/verify` and `POST /api/domains/:domain/delete` in an encapsulated Fastify scope with discard parsers for `application/x-www-form-urlencoded` and `*`. Same pattern as PR #477 (sign-out fix).

HTML form submits send `Content-Type: application/x-www-form-urlencoded`. Fastify had no registered parser for this → 415 before the handler ran. Neither endpoint reads the request body; they only use URL params and the session cookie.

Fixes #481.

## Test plan

- [x] `cd apps/api && bun run test --run` — 147 tests pass
- [ ] Manual: Sign in → `/dashboard/domains` → click "Re-verify" on any domain → returns 200/verified result (not 415)
- [ ] Manual: Click "Remove" on a domain → 302 redirect back to `/dashboard/domains`, domain disappears from list
- [ ] Manual: Screenshot the Re-verify and Remove actions as evidence

### DIFF
--- a/apps/api/src/routes/domains.ts
+++ b/apps/api/src/routes/domains.ts
@@ -325,91 +325,127 @@ export async function registerDomainsRoutes(
 
   // ──────────────────────────────────────────────────────────────────────────
   // POST /api/domains/:domain/verify — probe & mark verified on first match.
+  // POST /api/domains/:domain/delete — form-submit fallback (issue #83).
+  //
+  // Both routes receive HTML form POSTs (Content-Type:
+  // application/x-www-form-urlencoded). Fastify has no default parser for
+  // that content type, so without a registered parser it returns 415 before
+  // the handler runs. Neither route reads the request body — they only use
+  // URL params and the session cookie — so we register discard parsers in an
+  // encapsulated scope to keep them from bleeding into other routes.
+  // Same pattern as the sign-out route in auth.ts (fixes #481).
   // ──────────────────────────────────────────────────────────────────────────
-  app.post<{ Params: { domain: string } }>(
-    '/api/domains/:domain/verify',
-    { preHandler: [sessionMw] },
-    async (req, reply) => {
-      const account = req.account!;
-      const rawDomain = req.params.domain;
-      const domain = normalizeEtldPlusOne(rawDomain);
-      if (!domain) {
-        return reply.code(400).send({ error: 'invalid domain' });
-      }
+  await app.register(async function domainFormScope(scope) {
+    // Accept application/x-www-form-urlencoded — discard the body.
+    scope.addContentTypeParser(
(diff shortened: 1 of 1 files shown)


NikolayS:
VERDICT: APPROVE

The encapsulated scope is correctly applied. Both parsers (`application/x-www-form-urlencoded` and `*`) are registered on `scope`, not `app`, so they cannot bleed to other routes. Both `/verify` and `/delete` are registered inside the scope closure. The existing routes (`GET /api/domains`, `POST /api/domains`, `DELETE /api/domains/:domain`) remain on `app` directly — no regression.

One minor observation: the `*` wildcard parser is the same as in the auth.ts sign-out fix (PR #477) and is isolated to the scope, so it is safe here. No action needed.

Clean fix, consistent with the established pattern.

PR #490, opened 2026-04-27
infra: add systemd services for visitor-worker and aggregator (fixes #489)

NikolayS:
## Summary

No systemd services existed for the visitor worker or the aggregator. New studies were permanently stuck in `visiting` status — the visitor worker and aggregator must run as persistent processes for studies to complete.

- `willbuy-visitor-worker.service` — runs `scripts/run-visitor-worker-dogfood.ts` polling loop continuously (mirrors willbuy-api.service pattern)
- `willbuy-aggregator-trigger.service` + `willbuy-aggregator-trigger.sh` — polls `studies WHERE status='aggregating'` every 30s and runs `docker run willbuy-aggregator --study-id <id>`

Fixes #489.

## Deploy steps

```sh
sudo cp /srv/willbuy/infra/systemd/willbuy-visitor-worker.service /etc/systemd/system/
sudo cp /srv/willbuy/infra/systemd/willbuy-aggregator-trigger.service /etc/systemd/system/
sudo cp /srv/willbuy/infra/systemd/willbuy-aggregator-trigger.sh /usr/local/bin/
sudo chmod +x /usr/local/bin/willbuy-aggregator-trigger.sh
# Build aggregator Docker image first:
cd /srv/willbuy/apps/aggregator && sudo docker build -t willbuy-aggregator .
sudo systemctl daemon-reload
sudo systemctl enable --now willbuy-visitor-worker willbuy-aggregator-trigger
```

## Test plan

- [ ] Manual: `sudo systemctl status willbuy-visitor-worker willbuy-aggregator-trigger` → both show Active: active (running)
- [ ] Manual: Create a NEW study from `/dashboard/studies/new` → wait 5–15 minutes → study transitions through `capturing → visiting → aggregating → ready`
- [ ] Manual: Screenshot study status page showing "Ready" as evidence that the full pipeline completed

PR #495, opened 2026-04-27
infra: add willbuy-capture-worker.service (fixes #494)

### LINKED ISSUE #494: infra: add willbuy-capture-worker.service — studies stuck at capturing without it
## Problem

PR #490 added \`willbuy-visitor-worker.service\` (visiting → aggregating) and \`willbuy-aggregator-trigger.service\` (aggregating → ready/failed), but missed the first link in the chain: \`willbuy-capture-worker.service\`.

The full status pipeline:
1. POST /studies → \`capturing\` (API, done)
2. **capture-worker** polls visits where study.status='capturing', runs Playwright, pushes artifact to broker → transitions study to \`visiting\` ← **NO SERVICE**
3. visitor-worker polls where study.status='visiting' → LLM scoring → \`aggregating\` ← PR #490 ✅
4. aggregator-trigger docker run → \`ready\`/\`failed\` ← PR #490 ✅

Without a service unit for the capture-worker, every study created after deploy will permanently stall at \`capturing\`.

## Entry point

\`apps/capture-worker/src/index.ts\` — when run directly via \`bun run src/index.ts\` it starts the polling loop (reads \`DATABASE_URL\` + \`BROKER_SOCKET_PATH\`).

## Required

Add \`infra/systemd/willbuy-capture-worker.service\` modelled on \`willbuy-visitor-worker.service\`:
- \`User=willbuy\`
- \`WorkingDirectory=/srv/willbuy\`
- \`EnvironmentFile=/etc/willbuy/app.env\`
- \`ExecStart=/home/willbuy/.bun/bin/bun run apps/capture-worker/src/index.ts\`
- \`After=network.target willbuy-api.service willbuy-capture-broker.service\`
- \`Restart=on-failure\`

Deploy step: \`sudo systemctl enable --now willbuy-capture-worker\` (same as the other workers in issue #488).

## Acceptance

- Service file in \`infra/systemd/\`
- Smoke evidence: after deploy, \`journalctl -u willbuy-capture-worker -n 20\` shows the worker starting and polling; a test study created via POST /api/studies transitions to \`visiting\` within ~60s

## Context

Discovered during pre-deploy review of issue #489 / PR #490.

NikolayS:
## Summary

Adds `infra/systemd/willbuy-capture-worker.service`, the missing stage-1 unit for the study pipeline. PR #490 added services for stages 2 (visitor-worker) and 3 (aggregator-trigger) but omitted stage 1; this closes that gap so the full `capturing → visiting → aggregating → ready/failed` pipeline is covered by systemd units.

## Deploy step

```
sudo systemctl enable --now willbuy-capture-worker
```

## Smoke test after deploy

```
journalctl -u willbuy-capture-worker -n 20
```

Expected: log lines showing the worker polling for studies in `capturing` state.

Closes #494

### DIFF
--- a/infra/systemd/willbuy-capture-worker.service
+++ b/infra/systemd/willbuy-capture-worker.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=willbuy capture worker — polls DB for studies in capturing state, runs Playwright, transitions to visiting
+After=network.target willbuy-api.service willbuy-capture-broker.service
+
+[Service]
+Type=simple
+User=willbuy
+WorkingDirectory=/srv/willbuy
+EnvironmentFile=/etc/willbuy/app.env
+ExecStart=/home/willbuy/.bun/bin/bun run apps/capture-worker/src/index.ts
+Restart=on-failure
+RestartSec=3
+
+[Install]
+WantedBy=multi-user.target
--- a/scripts/smoke-test.sh
+++ b/scripts/smoke-test.sh
@@ -57,12 +57,17 @@ http_get_with_body() {
 
 # ──────────────────────────────────────────────
 # Test 2 — Demo report fixture
-# GET /r/test-fixture → 200
+# GET /r/test-fixture → 200 + body contains report content (not "Report not found")
+# Next.js returns 200 even for the "not found" error page, so check body too.
 # ──────────────────────────────────────────────
 {
-  status=$(http_get "${BASE_URL}/r/test-fixture")
-  if [[ "$status" == "200" ]]; then
-    pass "2. Demo report (GET /r/test-fixture → 200)"
+  tmp=$(mktemp)
+  status=$(curl -s --max-time 10 -w "%{http_code}" -o "$tmp" "${BASE_URL}/r/test-fixture" 2>/dev/null) || status="000"
+  body=$(cat "$tmp"); rm -f "$tmp"
+  if [[ "$status" == "200" ]] && ! echo "$body" | grep -q "Report not found"; then
+    pass "2. Demo report (GET /r/test-fixture → 200 + fixture rendered)"
+  elif [[ "$status" == "200" ]] && echo "$body" | grep -q "Report not found"; then
+    fail "2. Demo report" "got 200 but page shows 'Report not found' — run next build to deploy fixture"
   else
     fail "2. Demo report" "expected 200, got $status"
   fi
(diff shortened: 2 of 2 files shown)


NikolayS:
## REV review — PR #495 infra: add willbuy-capture-worker.service

**Reviewed files:** `infra/systemd/willbuy-capture-worker.service` (primary target), plus `apps/api/src/routes/auth.ts`, `apps/api/test/auth.test.ts`, `scripts/smoke-test.sh` (also on this branch).

---

### Service file analysis

**`After=` dependency chain** — correctly lists `willbuy-capture-broker.service`. The capture worker connects to the broker socket (`BROKER_SOCKET_PATH`), so this ordering dependency is required and present. ✓

**`User=willbuy`** — matches `willbuy-api.service` and `willbuy-web.service`. The broker uses `User=root` (it needs to create the socket and set `Group=willbuy`), which is a deliberate divergence; the worker only reads from the socket and runs Playwright under an unprivileged account. ✓

**`WorkingDirectory=/srv/willbuy`** — identical to `willbuy-api.service` and `willbuy-web.service`. ✓

**`EnvironmentFile=/etc/willbuy/app.env`** — matches the API and web services. Note: the broker uses `/etc/willbuy/broker.env` with the `-` prefix (optional). The worker correctly reads from `app.env`, which is where `DATABASE_URL` and `BROKER_SOCKET_PATH` live for other app-layer workers. ✓

**`ExecStart`** — `bun run apps/capture-worker/src/index.ts` is correct. Confirmed in `apps/capture-worker/src/index.ts`: the file has a guarded `isMain` block that starts the polling loop only when run directly (`process.argv[1]` check). The polling loop starts via `runPollingLoop()`, reads `DATABASE_URL` and `BROKER_SOCKET_PATH` from env, and handles `SIGTERM` with a 60 s drain window. Entry point is sound. ✓

**`Restart=on-failure` / `RestartSec=3`** — consistent with API and web units (`RestartSec=3`). The broker uses `RestartSec=2`, a minor divergence that is not a concern here. ✓

**`Type=simple`** — correct; `bun run` does not fork. ✓

**`WantedBy=multi-user.target`** — standard. ✓

---

### Observations / non-blocking notes

1. **No hardening directives.** `willbuy-capture-broker.service` has a full hardening block (`NoNewPrivileges`, `ProtectSystem`, `PrivateTmp`, etc.). The capture worker runs Playwright, which spawns browser sub-processes and likely needs broader access, so some directives (e.g. `MemoryDenyWriteExecute`, `RestrictNamespaces`) would conflict with Playwright's sandboxing. Omitting hardening here is therefore defensible. A follow-up issue to audit which directives _are_ safe to add would be worthwhile, but is not a blocker.

2. **Other changes on this branch** (`apps/api/src/routes/auth.ts`, `apps/api/test/auth.test.ts`, `scripts/smoke-test.sh`) are unrelated to the stated scope of this PR (fixes #476 and a new smoke-test script). They look correct in isolation but should ideally travel in their own PR to keep history clean. Not a blocker given they appear already reviewed elsewhere, but worth noting for future hygiene.

3. **`After=willbuy-api.service`** is present but the worker has no direct dependency on the API (it talks to the DB and broker socket). This is harmless — the API starts quickly and the DB connection is independent — but it adds an unnecessary ordering constraint. Again, not a blocker.

---

## VERDICT: APPROVE

The service file is well-formed, the `After=` broker dependency is correct, User/WorkingDirectory/EnvironmentFile are consistent with the rest of the fleet, and the `ExecStart` entry point is verified against source. Non-blocking notes above can be addressed in follow-ups.

PR #504, opened 2026-04-27
fix(infra): add --network=host to aggregator docker run (DB connectivity)

NikolayS:
## Problem

The aggregator trigger script runs `docker run ... willbuy-aggregator --study-id <id>` without `--network=host`. On a single-server deployment, if `DATABASE_URL` uses `localhost` or a Unix socket, the container cannot reach the host's Postgres — it gets `connection refused`, the aggregator exits non-zero, and studies are permanently stuck at `aggregating`.

## Fix

Add `--network=host` so the container shares the host network stack. `localhost` inside the container then resolves to the host, matching the `DATABASE_URL` format used by the other services.

This is appropriate for the single-server MVP. Multi-host deployments would need a proper overlay network.

## Test

After deploy: create a study, watch it reach `aggregating`, confirm `docker run` in `journalctl -u willbuy-aggregator-trigger` exits 0 and study transitions to `ready`.

### DIFF
--- a/infra/systemd/willbuy-aggregator-trigger.sh
+++ b/infra/systemd/willbuy-aggregator-trigger.sh
@@ -25,6 +25,7 @@ while true; do
   if [[ -n "$study_id" ]]; then
     echo "[aggregator-trigger] running aggregator for study $study_id"
     docker run --rm \
+      --network=host \
       --env DATABASE_URL="$DATABASE_URL" \
       willbuy-aggregator \
       --study-id "$study_id"


NikolayS:
## VERDICT: APPROVE

One-line infra fix.  is the correct solution for a single-server deployment where `DATABASE_URL` uses `localhost`. Without it the aggregator container always fails to connect to Postgres and studies never reach `ready`. Non-blocking note: for future multi-host setups this would need an overlay network instead, but that's out of scope for the v0.1 single-server MVP.

